PR #19, opened 2015-01-11
Fix issue with mro in zodb

do3cc:
@mgedmin Thanks for point me to the failing zodb tests.
I am not happy with this solution, but checking for the __mro__ attribute failed, probably because of the python trick to make __ private, and isinstance(layer, type) also failed, maybe because of old style layer classes.
I was also wondering, if it is legal to have an object as a layer, but the docs don't say.
I manually tests zodb with 3.4 and this change, and my tests then ran through

### DIFF
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -886,11 +886,15 @@ def order_by_bases(layers):
     """Order the layers from least to most specific (bottom to top)
        Group layers with common bases and put unittests first.
     """
-    getmro = inspect.getmro
+    def getmro(layer):
+        try:
+            return inspect.getmro(layer)
+        except AttributeError:
+            return inspect.getmro(layer.__class__)
 
     def layer_sortkey(layer):
         return tuple((c.__module__, c.__name__) for c in getmro(layer)[::-1]
-                      if c not in (object, UnitTests))
+                     if c not in (object, UnitTests))
 
     layers = sorted(layers, key=layer_sortkey, reverse=True)
     gathered = []


mgedmin:
Layers don't have to be classes.  They're objects with a `__bases__` attribute listing base layers.  I've been confused by this in the past.

1. We should have an explicit test of this use-case.

2. I fear `layer.__class__.__mro__` is wrong and we need to do a manual traversal of `layer.__bases__`.

do3cc:
I fear it must be both classes and __mro__
I fear I fell into a deep pit

do3cc:
I have reviewed code and docs a bit more. zope.testrunner keeps it simple for itself.
They do not state if a layer has to be a class, but they state that Layers can be extended giving an example of extending classes via subclassing. http://docs.zope.org/zope.testrunner/index.html#layers
In the implementation, base classes are looked up by going over the elements of the `__bases__` property which is a python property. https://docs.python.org/2/library/stdtypes.html#special-attributes
Since it is not part of the zope.testrunner docs, I would say that it is reasonable to expect a layer to be a class like attribute. If one wants to fake it, depending on internal knowledge of how resolution works, one has to suck it up that a new patch release changes internal details and one has to provide a __mro__ method too now.
Sorting by mro is safe because layer setup is done breath-first. __mro__ works the same, mostly.

mgedmin:
[zope.testrunner's documentation](https://github.com/zopefoundation/zope.testrunner/blob/673f70cdcf04a5333b1adf44c8922ae32c850f8f/src/zope/testrunner/tests/testrunner-layers-api.txt#L9) says "Layers are *generally* implemented as classes using class methods" (emphasis mine), which can be read to imply that they don't have to be.

Martin Aspeli describes some downsides of using classes themselves as test layers in this [email from 2010](https://mail.zope.org/pipermail/zope-dev/2010-April/040267.html).

Some market research:

- [ZODB](https://github.com/zopefoundation/ZODB/blob/adf92861902125c7ca70854d9354557806d45217/src/ZODB/tests/util.py#L99-L118) uses an object instance as the layer.
- [zope.app.testing.functional](https://github.com/zopefoundation/zope.app.testing/blob/master/src/zope/app/testing/functional.py#L320-L359), uses an object instance as the layer.
- [zope.app.wsgi](https://github.com/zopefoundation/zope.app.wsgi/blob/0a50583193daf85d16d81731a9b1358d8074e3f4/src/zope/app/wsgi/tests.py#L74-L94) uses object instances as the layer.
- [plone.testing](https://github.com/plone/plone.testing/blob/master/src/plone/testing/layer.py) defines object instances to be used as test layers.
- [gocept.selenium](https://bitbucket.org/gocept/gocept.selenium/src/aaa8166f5244c3fcc4506d17ffbe6166bafac5cf/src/gocept/selenium/webdriver.py?at=default#cl-33) builds on top of plone.testing and therefore also uses object instances as test layers.

In short, I don't think zope.testrunner can afford to drop support of object instances as test layers (or require that everyone go implement a `__mro__` property on their test layers) at this time, without a painful transition affecting many users.

I suggest that reimplementing the MRO algorithm inside zope.testrunner to compute the MRO from `layer.__bases__` is the right thing to do.  I volunteer to do it if nobody else steps forward. ;)

mgedmin:
Welp, repozo --verify is broken in ZODB 4.1.0.

mgedmin:
Actually, my original --verify patch was broken.  The -V spelling works.

tseaver:
@mgedmin what does repozo have to do with this issue (or even this repository)?

mgedmin:
Um, how did these messages end up in GitHub?  I think I meant to type them into Twitter.

Let's blame focus-follows-mouse and late evening tiredness.

mgedmin:
What does repozo do with the last incomplete transaction?  In theory we have a race:

1. A new transaction starts appending to Data.fs
2. ls -l Data.fs > last-backup.txt
3. repozo runs, finds an incomplete transaction at the end of the file, seeks backwards to the beginning of the transaction
4. rsync

and now last-backup.txt says we should have more bytes than we actually have, even though nothing is wrong.

TODO:

- [ ] read repozo source code to see if it really seeks backwards.

mgedmin:
Welp, somebody please take my computer away from me.  I keep typing comments into wrong browser tabs.  :(

do3cc:
Just for the record, I volunteered too. For reverting 4.4.5...

mgedmin:
Oh my, this is a can of worms.

I'm now worrying about the implications of zope.testrunner not using the C3 MRO algorithm when running layer setup.

(AFAICS the optimal layer ordering solution that minimizes the total number of layer setUp/tearDown calls is NP-hard.)

mgedmin:
#20 is my attempt to fix this by moving forwards rather than backwards.  Review would be appreciated!

mgedmin:
Fixed by #20.

Issue #6, opened 2013-05-29
zope-testrunner -j99 fails when run from tox


mgedmin:
ZEO has a test suite with multiple layers.  The tests are hooked up with bin/test and with tox.  Both use zope.testrunner.  bin/test -j99 works great, but if you configure tox to run zope-testrunner -j99, you get a bunch of errors like this:

```
**********************************************************************
Could not communicate with subprocess!
Child stderr was:
      /full/path/to/.tox/py27/bin/python: can't open file '.tox/py27/bin/zope-testrunner': [Errno 2] No such file or directory
**********************************************************************
```

PR #20, opened 2015-01-14
Fix instance based layers

mgedmin:
Commit 3a68ee9c41f4da4f6f64a1fb3fa2801a9a8e322b improved layer sorting but accidentally dropped support for instance-based layers.  These turn out to be rather widely used (in ZODB, in zope.app.testing, in plone.testing etc).

This change fixes the above problem in a different way from PR #6: it replacing inspect.getmro(layer) with a recursive sort key computation that traverses layer.\__bases__.

It doesn't actually use the C3 algorithm for MRO computation.  I tried to find an example where that would matter and couldn't.

This change also adds tests to prevent this from regressing in the future and defines the test runner's expectations for test layers as interfaces.

### DIFF
--- a/src/zope/testrunner/interfaces.py
+++ b/src/zope/testrunner/interfaces.py
@@ -100,3 +100,60 @@ class ITestRunner(zope.interface.Interface):
 
     options = zope.interface.Attribute(
       "Provides access to configuration options.")
+
+
+class IMinimalTestLayer(zope.interface.Interface):
+    """A test layer.
+
+    This is the bare minimum that a ``layer`` attribute on a test suite
+    should provide.
+    """
+
+    __bases__ = zope.interface.Attribute(
+        "A tuple of base layers.")
+
+    __name__ = zope.interface.Attribute(
+        "Name of the layer")
+
+    __module__ = zope.interface.Attribute(
+        "Dotted name of the module that defines this layer")
+
+    def __hash__():
+        """A test layer must be hashable"""
+
+
+class IFullTestLayer(IMinimalTestLayer):
+    """A test layer.
+
+    This is the full list of optional methods that a test layer can specify.
+    """
+
+    def setUp():
+        """Shared layer setup.
+
+        Called once before any of the tests in this layer (or sublayers)
+        are run.
+        """
+
+    def tearDown():
+        """Shared layer teardown.
+
+        Called once after all the tests in this layer (and sublayers)
+        are run.
+
+        May raise NotImplementedError.
+        """
+
+    def testSetUp():
+        """Additional test setup.
+
+        Called once before every of test in this layer (or sublayers)
+        is run.
+        """
+
+    def testTearDown():
+        """Additional test teardown.
+
+        Called once after every of test in this layer (or sublayers)
+        is run.
+        """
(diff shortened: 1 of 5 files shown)


src/zope/testrunner/interfaces.py
```diff
@@ -100,3 +100,60 @@ class ITestRunner(zope.interface.Interface):
...
+        "A tuple of base layers.")
+
+    __name__ = zope.interface.Attribute(
+        "Name of the layer")
+
+    __module__ = zope.interface.Attribute(
+        "Dotted name of the module that defines this layer")
+
+    def __hash__():
+        """A test layer must be hashable"""
```
tseaver: Does this need an `__eq__` as well?  The two pretty much have to be defined using the same semantics, or else things break.

src/zope/testrunner/tests/testrunner-layers-instances.txt
```diff
@@ -0,0 +1,183 @@
+Layers implemented via object instances
+=======================================
+
+Layers are generally implemented as classes using class methods, but
+regular objects can be used as well.  They need to provide __module__,
+__name__, and __bases__ attributes.
+
+>>> class TestLayer:
```
tseaver: Should derive from `object` (for clarity, because we straddle Python2 / Python3).

tseaver: Should we declare that we implement `IFullTestLayer`?  (I realize the runner doesn't check it)

src/zope/testrunner/interfaces.py
mgedmin: I wasn't sure if I should include \__eq__ in the interface:

- On one hand, to be properly hashable an object has to have \__eq__ and \__hash__ and the two have to be consistent with each other.

- On the other hand Python object instances (including classes themselves) are identity-comparable and hashable by default, so the user doesn't have to do anything (unless they decide to override \__eq__).

So I chose a compromise and documented one of the two, thus combining the downsides.

I think you're right, a reminder about \__eq__ would be good.  (And also a comment that the default implementations provided by Python suffice.)

src/zope/testrunner/tests/testrunner-layers-instances.txt
mgedmin: > Should derive from object (for clarity, because we straddle Python2 / Python3).

I've seen [layers in the wild](https://github.com/zopefoundation/zope.app.testing/blob/ac2353c932bd055d0f8abcc2a74806f03d2f48ef/src/zope/app/testing/functional.py#L320) that are old-style classes on Python 2.  I wanted to be sure they continue to work.

> Should we declare that we implement IFullTestLayer? (I realize the runner doesn't check it)

Again, existing real-world usage doesn't do that and I wanted this test to be a representative example.

(Also, declaring that TestLayer is an `@implementer(IFullTestLayer)` would raise questions about doing the analoguous `@provider(...)` for class-based layers defined in the other tests.)

src/zope/testrunner/interfaces.py
mgedmin: I tried to add \__eq__ and it felt wrong to me again.

I don't oppose documenting that, but I can't find the words myself.

tseaver: I don't really care which, but we should either document both `__hash__` and `__eq__`, or neither.

tseaver:
LGTM

PR #23, opened 2015-03-17
Update python-subunit to 1.0.

tseaver:
Except, pin to '<1.0dev' for Python 2.6.

Note that tests fail on Py3k:  I'm pushing this as a PR to ask for
help from @rbtcollins

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,6 @@ python:
 install:
     - pip install .
 script:
-    - python setup.py test -q
+    - python setup.py -q test -q
 notifications:
     email: false
--- a/setup.py
+++ b/setup.py
@@ -30,17 +30,14 @@ if sys.version_info < (2,6) or sys.version_info[:2] == (3,0):
 
 
 
-if sys.version_info >= (3,):
-    tests_require = ['zope.testing']
-    extra = dict(# XXX:  python-subunit is not yet ported to Python3.
-                 extras_require = {'test': ['zope.testing']},
-                 )
+tests_require = ['zope.testing']
+if sys.version_info[0:2] == (2, 6):
+    tests_require.append('unittest2')
+    tests_require.append('python-subunit < 1.0dev')
 else:
-    tests_require = ['zope.testing', 'python-subunit']
-    if sys.version_info[0:2] == (2, 6):
-        tests_require.append('unittest2')
-    extra = dict(tests_require = tests_require,
-                 extras_require = {'test': tests_require})
+    tests_require.append('python-subunit')
+extra = dict(tests_require = tests_require,
+                extras_require = {'test': tests_require})
 
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -8,21 +8,12 @@ envlist =
 [testenv]
 deps =
     zope.testing
+    python-subunit
 commands =
-    python setup.py test -q
+    python setup.py -q test -q
 
 [testenv:py26]
 deps =
-    {[testenv]deps}
-    python-subunit
+    zope.testing
+    python-subunit<1.0dev
     unittest2
-
-[testenv:py27]
-deps =
-    {[testenv]deps}
-    python-subunit
-
-[testenv:pypy]
-deps =
-    {[testenv]deps}
-    python-subunit


tseaver:
@rbtcollins, can you help with debugging the Py3k failures?   See https://gist.github.com/tseaver/4e968b4eb2d3e77645cb

rbtcollins:
Sure, let me have a look.

rbtcollins:
Ok, your SpoofOut appears to be the issue - AttributeError: '_SpoofOut' object has no attribute 'buffer' - https://github.com/testing-cabal/subunit/blob/master/python/subunit/__init__.py#L1288-L1304 - if subunit reads unicode from a stream (it needs binary in and out), it unwraps it to make it into a binary stream, by grabbing the underlying buffer attribute. Since Py2.xisn't binary aware by default, there isn't a 2.x internals unwrapping codepath there - and for testing I use io's structure (works on 2.6 too) - https://github.com/testing-cabal/subunit/blob/master/python/subunit/tests/test_run.py#L100-L101

rbtcollins:
(That is, using the io classes should work for you for 2.6 all the way through 3.5, and is the stock structure for streams, so should work with a broad spread of code).

rbtcollins:
Oh one more note - python-subunit is still tested and supported on 2.6, you shouldn't need to pin it differently.

tseaver:
@rbtcollins thanks for looking into it.  The Python 2.6 must've been me getting things into a weird state:  `tox --recreate -e py26` fixed it after backing out the pin.

WRT the other failures:  have a look at the ones on the end, where the stream is an instance of `io.StringIO`, rather than `_SpoofOut`.  Should it be part of the contract of `zope.testrunner.formatter.SubunitOutputFormatter` that the stream is binary, rather than text?  It falls back to `sys.stdout`, which is text (I think).

tseaver:
```python
$ .tox/py34/bin/python
Python 3.4.2 (default, Feb 13 2015, 11:28:12) 
[GCC 4.6.3] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import io
>>> x = io.StringIO()
>>> from subunit import _unwrap_text
>>> _unwrap_text(x)
  File "<stdin>", line 1, in <module>
  File "...tox/py34/lib/python3.4/site-packages/subunit/__init__.py", line 1297, in _unwrap_text
    return stream.buffer
AttributeError: '_io.StringIO' object has no attribute 'buffer'
```

tseaver:
Also, `_SpoofOut` is defined in the stdlib's `doctest` module:  it derives directly from `io.StringIO`.

rbtcollins:
Right - StringIO directly implements read and write of unicode - thats why the code I linked to uses BytesIO + a TextIOWrapper for testing.

Subunit v1 and v2 are both bytes orientated so yes, passing in BytesIO in the first place should work just fine. If the Zope layer code has a String stream thats fine, as long as subunit has appropriate code to unwrap it - all the failures looked to be in the unwrap code, but all due to test code not being equivalent to the io layering (stream either being bytes itself, or stream.buffer existing and being a bytes file).

rbtcollins:
Seems to me that the smallest change would be to make the tests use io.TextIOWrapper + BytesIO rather than StringIO. Changing the contract to require a Bytes stream for the zope formatter would be more complex, and lead to skew between the different formatters (or awkward code in text orientated formatters.

tseaver:
I could (maybe) fix the three unit test cases that way:  the problem is that the stdlib's `doctest` module creates the `_SpoofOut` class it uses to dummy output.  Shouldn't `subunit` work with doctests on Py3k?  And if so, wouldn't it be best to have `subunit._unwrap_test` do the Right Thing(TM) with instances of `io.StringIO`, too?

rbtcollins:
I'm not sure there is a Right Thing with io.StringIO.

Here's the situation:
subunit needs to write bytes (actual binary, not utf8 encoded text).
io.StringIO() has no bytes backing buffer, so you have to take the bytes, do something to them to make them unicode and then you can write them to the StringIO object.

We could decode with one of the codecs that can't fail like cp926 or whatever the magic # is. That seems pretty awful to put in the production code though, because its never going to be the right thing when outputting a stream - we care about the exact bytes hitting the network socket / file / inter-process pipe, and in that situation the unicode that we write to the stream would be getting re-encoded by something, potentially utf8 or whatever, and things would be entirely broken in that situation.

What are you doctesting with subunit? ... goes and looks in detail - ok - https://github.com/zopefoundation/zope.testrunner/blob/master/src/zope/testrunner/tests/testrunner-subunit.txt#L44

Thats taking a subunit v1 stream, a *subset* of which are US-ASCII (but not all of) and using that as a way to check the output. But its not representative of the actual way subunit streams behave in the wild - they are line ending sensitive, for instance, so a text stream with different EOL markers will break parsing the stream. OTOH for the specific inputs you're giving it its probably pragmatically ok.

Perhaps monkey patch SpoofOut in the interim (to make https://github.com/testing-cabal/subunit/blob/master/python/subunit/__init__.py#L1296 pass - give it a read(0) method which returns b'').

Then we can talk about the best way to test your subunit support (and also support subunit-v2 which is  much more capable - native multiplexing and concurrency for instance) without the time pressure of a failing test suite.

-Rob

tseaver:
@rbtcollins At this point I'm preparing a PR to drop all support for `subunit`:  solid Py3k testing support is way more important to us.

If you'd like us to continue supporting it, please push a PR which updates `zope.testrunner` to latest-and-greatest subunit and passes all tests on all supported Python versions.

PR #25, opened 2015-04-30
Support skipped tests in subunit output

kedder:
Previously, attempting to run test suite, containing skipped tests with
``--subunit`` option produced exception:

    AttributeError: 'SubunitOutputFormatter' object has no attribute 'test_skipped'

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@ zope.testrunner Changelog
 4.4.8 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Support skipped tests in subunit output
 
 
 4.4.7 (2015-04-02)
--- a/src/zope/testrunner/formatter.py
+++ b/src/zope/testrunner/formatter.py
@@ -959,6 +959,9 @@ class SubunitOutputFormatter(object):
         self._emit_timestamp()
         self._subunit.addSuccess(test)
 
+    def test_skipped(self, test, reason):
+        self._subunit.addSkip(test, reason)
+
     def import_errors(self, import_errors):
         """Report test-module import errors (if any)."""
         if not import_errors:
--- a/src/zope/testrunner/tests/testrunner-subunit.txt
+++ b/src/zope/testrunner/tests/testrunner-subunit.txt
@@ -675,6 +675,39 @@ Note that debugging doesn't work when running tests in a subprocess:
     ]
     True
 
+
+Support skipped tests
+---------------------
+
+    >>> directory_with_skipped_tests = os.path.join(this_directory,
+    ...                                             'testrunner-ex-skip')
+    >>> skip_defaults = [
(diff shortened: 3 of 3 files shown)


mgedmin:
Travis fails all Python 3 jobs:
> Subunit is not installed. Please install Subunit
> to generate subunit output.

I see subunit [claims to support Python 3](https://pypi.python.org/pypi/python-subunit), so can you remove the [condition  in setup.py](https://github.com/zopefoundation/zope.testrunner/blob/62fb35421dc889c82332da0d9661e771e8cfa86d/setup.py#L33-L38)?

kedder:
I tried to include subunit into python-3 builds as well, but was met with exceptions like one below. So, instead I moved the test to ``testrunner-subunit.txt``. Which might seem opportunistic, but these python3/subunit problems does not seem to be related to this issue directly.

```
Error in test test_print_failure_containing_utf8_bytestrings (zope.testrunner.tests.test_subunit.TestSubunitTracebackPrinting)
  File "/usr/lib/python3.4/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib/python3.4/unittest/case.py", line 573, in run
    self.setUp()
  File "/home/andrey/projects/shoobx/zope.testrunner/.tox/py34/lib/python3.4/site-packages/zope/testrunner/tests/test_subunit.py", line 57, in setUp
    options, stream=self.output)
  File "/home/andrey/projects/shoobx/zope.testrunner/.tox/py34/lib/python3.4/site-packages/zope/testrunner/formatter.py", line 795, in __init__
    self._subunit = subunit.TestProtocolClient(self._stream)
  File "/home/andrey/projects/shoobx/zope.testrunner/.tox/py34/lib/python3.4/site-packages/subunit/__init__.py", line 621, in __init__
    stream = make_stream_binary(stream)
  File "/home/andrey/projects/shoobx/zope.testrunner/.tox/py34/lib/python3.4/site-packages/subunit/__init__.py", line 1278, in make_stream_binary
    return _unwrap_text(stream)
  File "/home/andrey/projects/shoobx/zope.testrunner/.tox/py34/lib/python3.4/site-packages/subunit/__init__.py", line 1297, in _unwrap_text
    return stream.buffer
AttributeError: '_io.StringIO' object has no attribute 'buffer'
```

kedder:
BTW, subunit py3 issues are discussed in #23.

mgedmin:
+1 for disabling the new subunit test on Python 3 then.

mgedmin:
Please ping me if you want a new zope.testrunner release on PyPI, or if you want PyPI rights so you could make the release.

kedder:
Alright. I think it will need another patch before newer zope.testrunner will be useful for us -- it has some sort of regression that makes it take several minutes to calculate tests to run, when there's a lot (e.g. 500) of ``-t`` parameters on command line. But I have no details yet, so this is not a proper bug report.

Issue #28, opened 2015-05-19
make -jN work properly


mgedmin:
Currently zope-testrunner -jN runs the first test layer serially, and only then starts forking N subprocesses for parallel execution.  People have been creating workarounds for this by adding an empty test layer named something like `.!no tests here!` to get the 1st serial layer out of the way ASAP, so the rest of the tests can run in parallel.

The change in 4.4.5 (run the unit test layer first, https://bugs.launchpad.net/zope.testrunner/+bug/497871) broke this hack.

Suggestion: if the test runner sees -j on the command line, let it create an empty test layer automatically and insert it in front of all other layers.  (Maybe make it conditional on there being at least two other layers, maybe don't bother -- who would use -j when there's only one test layer?)

mgedmin:
BTW fixing this will make the regression test for #6 invalid.  The test should probably be rewritten to avoid -j and instead explicitly define a test layer that does os.chdir() and raises NotImplementedError from layer teardown.

mgedmin:
Released in 4.4.9.

PR #29, opened 2015-05-19
fixing #28, auto add a fake layer when -j is used

agroszer:


tseaver:
Who would've thought there would be race conditions in a doctest of parallel test suites?  OH, NOES!

src/zope/testrunner/layer.py
```diff
@@ -13,7 +13,26 @@
...
 """
+import unittest
 
 
 class UnitTests(object):
     """A layer for gathering all unit tests."""
+
+
+class FakeMultiProcessLayer(object):
+    """A fake layer to start spreading out subprocesses."""
```
mgedmin: How about "EmptyLayer"?  There's nothing fake about it ("FakeFoo" is what I name my stubs for use in unit tests; this is a real thing).

The docstring could say "A layer with no tests so subprocesses can start at once" or something like that.

src/zope/testrunner/layer.py
```diff
@@ -13,7 +13,26 @@
...
+
+class FakeMultiProcessTests(unittest.TestCase):
+    """A fake TestCase to start spreading out subprocesses."""
+    layer = FakeMultiProcessLayer()
+
+
+def FakeMultiProcessTests_suite():
+    return unittest.TestSuite((
+        unittest.makeSuite(FakeMultiProcessTests),
+    ))
```
mgedmin: This seems needlessly complicated.

```python
def  EmptySuite():
    suite = unittest.TestSuite()
    suite.layer = EmptyLayer()
    return suite
```

should suffice.

src/zope/testrunner/runner.py
```diff
@@ -125,6 +127,13 @@ def __init__(self, defaults=None, args=None, found_suites=None,
         self.tests_by_layer_name = {}
 
     def ordered_layers(self):
+        if (self.options.processes > 1 and not self.options.resume_layer):
+            # if we want multiple processes, we need a fake layer as first
+            # to start spreading out layers/tests to subprocesses
+            # but only if this is not in the subprocess
+            yield (name_from_layer(FakeMultiProcessLayer),
+                   FakeMultiProcessLayer, FakeMultiProcessTests_suite())
+
```
mgedmin: I'm surprised that this doesn't have to be added for subprocesses, but as long as it works.

src/zope/testrunner/tests/testrunner-ex/sampletests_buffering.py
```diff
@@ -30,7 +30,7 @@ def tearDown(self):
 class Layer2:
 
     def setUp(self):
-        pass
+        time.sleep(0.1)
```
mgedmin: :(

Also, this appears to be [insufficient for PyPy3](https://travis-ci.org/zopefoundation/zope.testrunner/jobs/63174193), maybe they optimize `time.sleep()` too well ;)

mgedmin:
The rest LGTM, mostly.

src/zope/testrunner/tests/testrunner-layers-cwd.txt will need to be rewritten to be a useful regression test for #6, but I suppose I can do that later.

The non-determinism in tests for parallel execution is (a) new, and (b) bad.  I should be fixed before we merge this.  I see two ways: (1) make the execution deterministic, or (2) make the test ignore the execution order of the subproceses and accept both valid outputs.  (2) would probably require us to rewrite the test as a non-doctest.  Or do a custom doctest output checker.  What is the test actually trying to test?

src/zope/testrunner/runner.py
agroszer: Tried first without ```self.options.resume_layer``` but that ended up in an infinite loop. Every subprocess ran the FakeMultiProcessLayer tests, then spawn a subprocess for the real tests and so on.

mgedmin: Ah, I see now that `self.tests_by_layer_name` contains the *filtered* set of test layers to run.

Well, originally it contains the unfiltered set, and then [it gets filtered in-place](https://github.com/zopefoundation/zope.testrunner/blob/master/src/zope/testrunner/filter.py#L31-L53).

(Oh my, now I that I've seen the filtering code feel like I need to make sure unit tests don't get run repeatedly when subprocesses are involved.)

agroszer:
re non-determinism:
I think the test tries to verify that there's some timely output from the tests in the subprocess. IOW we need to show that something is happening. Could be reduced to the fact that there are three "." (dots) emitted?

Not much help, but I've digged up this: http://svn.zope.org/zope.testrunner/trunk/src/zope/testrunner/testrunner-layers-buff.txt?view=log&pathrev=123154

(a) Before Layer1 was run in the master process as first (instead of EmptyLayer) and Layer2 after. Now Layer1 and Layer2 are run parallel.
(1) Sounds hard, any ideas considering that those 2 setUp's are running in different processes?
(2) meeeh

mgedmin:
I'm not opposed to a longer time.sleep() in the setUp of *one* of the layers as a temporary solution, if that reduces the chances of a test failure.

I really want to get #28 fixed soon, but I won't merge something that fails on Travis.

mgedmin:
I tried a tox run locally, got a failure in testrunner-layers-buff.txt with pypy.

BTW I also discovered that tox --develop doesn't work (ImportError: No module named testrunner), which annoyingly makes this harder to debug.

mgedmin:
On a re-run py26 failed in testrunner-layers-buff.txt.

mgedmin:
Reviewed the history of testrunner-layers-buff:
- [current](https://github.com/zopefoundation/zope.testrunner/commits/master/src/zope/testrunner/tests/testrunner-layers-buff.txt)
- [before the move into tests/](https://github.com/zopefoundation/zope.testrunner/commits/cc277d77df6aa72969d31c23b1b64e1ebee4e090/src/zope/testrunner/testrunner-layers-buff.txt)
- [before the move out of zope.testing.testrunner](https://github.com/zopefoundation/zope.testing/commits/95592702de3af4656c9a67e4c98bcc9ffefb133d/src/zope/testing/testrunner/testrunner-layers-buff.txt)

Interesting commits:
- 83d25ac7be4e9c0435efe122e32d09fb4aaeb9da: this test is now (re)used as a regression test for #7 (0d4b2d6a49e40b36dd809b9c748e59e768f9f5f6 is a fixup for the same)
- de501b0e4e41833f11517903dd1d4f4ce9358888: a fix for sporadic failures
- https://github.com/zopefoundation/zope.testing/commit/da43ed5533b8ff0f61ff03fbaff43b981034fe9c: this test is now (re)used as a regression test for subprocess error reporting
- https://github.com/zopefoundation/zope.testing/commit/e9157548b1a9b5148a5632b78257c8c225f25081: an earlier attempt to make this test more reliably
- https://github.com/zopefoundation/zope.testing/commit/701f9911873f45cdbfde5a3409a04b93a35f2935: the commit that added this test

mgedmin:
BTW a reliable way for me to reproduce the non-deterministic failure was to run `detox` in one terminal and `bin/test -pvc -t layers-buff --repeat 50` in another one.

mgedmin:
Thank you for doing this!

Issue #30, opened 2015-11-03
1 Test failure with Python 3.5


felixonmars:
The error message `AttributeError: module 'traceback' has no attribute '_iter_chain'` points me to an upstream issue about Python 3.5. Hope this helps.

FYI: https://bugs.python.org/issue17911

```
Failure in test /build/python-zope-testrunner/src/zope.testrunner/src/zope/testrunner/tests/testrunner-debugging-layer-setup.test
Failed doctest test for testrunner-debugging-layer-setup.test
  File "/build/python-zope-testrunner/src/zope.testrunner/src/zope/testrunner/tests/testrunner-debugging-layer-setup.test", line 0

----------------------------------------------------------------------
File "/build/python-zope-testrunner/src/zope.testrunner/src/zope/testrunner/tests/testrunner-debugging-layer-setup.test", line 104, in testrunner-debugging-layer-setup.test
Failed example:
    try:
        zope.testrunner.run_internal(
          ['--path', dir, '-Dvv', '--tests-pattern', 'tests2'])
    except zope.testrunner.interfaces.EndRun: print("unexpected EndRun!")
    finally: sys.stdin = real_stdin
    # doctest: +ELLIPSIS +REPORT_NDIFF
Differences (ndiff with -expected +actual):
      Running tests at level 1
      Running tests2.Layer1 tests:
        Set up tests2.Layer1 in N.NNN seconds.
        Running:
       a_test (tests2)
        Ran 1 tests with 0 failures, 0 errors and 0 skipped in N.NNN seconds.
      Running tests2.Layer2 tests:
        Tear down tests2.Layer1 ... not supported
        Running in a subprocess.
    -   Set up tests2.Layer2
    +   Set up tests2.Layer2
    ?                       +
      **********************************************************************
      <BLANKLINE>
      Can't post-mortem debug when running a layer as a subprocess!
      Try running layer 'tests2.Layer2' by itself.
      <BLANKLINE>
      **********************************************************************
      <BLANKLINE>
      Traceback (most recent call last):
    - ...
    +   File "/build/python-zope-testrunner/src/zope.testrunner/src/zope/testrunner/runner.py", Line NNN, in run_layer
    +     setup_layer(options, layer, setup_layers)
    +   File "/build/python-zope-testrunner/src/zope.testrunner/src/zope/testrunner/runner.py", Line NNN, in setup_layer
    +     layer.setUp()
[...]

PR #31, opened 2015-11-07
Add support for Python 3.5

felixonmars:
This addresses #30

### DIFF
--- a/src/zope/testrunner/tb_format.py
+++ b/src/zope/testrunner/tb_format.py
@@ -20,8 +20,10 @@ import zope.exceptions.exceptionformatter
 import zope.testrunner.feature
 
 
-# Python 3.5 compatibility
-if not hasattr(traceback, "_iter_chain"):
+try:
+    _iter_chain = traceback._iter_chain
+except AttributeError:
+    # Python 3.5
     def _iter_chain(exc, custom_tb=None, seen=None):
         if seen is None:
             seen = set()
@@ -42,12 +44,10 @@ if not hasattr(traceback, "_iter_chain"):
             for x in it:
                 yield x
 
-    traceback._iter_chain = _iter_chain
-
 
 def format_exception(t, v, tb, limit=None, chain=None):
     if chain:
-        values = traceback._iter_chain(v, tb)
+        values = _iter_chain(v, tb)
     else:
         values = [(v, tb)]
     fmt = zope.exceptions.exceptionformatter.TextExceptionFormatter(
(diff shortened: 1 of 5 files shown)


src/zope/testrunner/tb_format.py
```diff
@@ -20,6 +20,31 @@
...
+            if context is not None and context not in seen:
+                its.append(_iter_chain(context, None, seen))
+                its.append([(traceback._context_message, None)])
+        its.append([(exc, custom_tb or exc.__traceback__)])
+        # itertools.chain is in an extension module and may be unavailable
+        for it in its:
+            for x in it:
+                yield x
+
+    traceback._iter_chain = _iter_chain
```
jamadden: Would it be better to avoid monkey-patching a stdlib module, and instead make `_iter_chain` a local function? Something like:
```python
try:
    _iter_chain = traceback._iter_chain
except AttributeError:
    # Python 3.5
    def _iter_chain(...):

...
def format_exception(...):
  if chain:
       values = _iter_chain(v, tb)
```

felixonmars: I see, updated :)

src/zope/testrunner/tb_format.py
```diff
@@ -20,9 +20,34 @@
...
+        else:
+            context = exc.__context__
+            if context is not None and context not in seen:
+                its.append(_iter_chain(context, None, seen))
+                its.append([(traceback._context_message, None)])
+        its.append([(exc, custom_tb or exc.__traceback__)])
+        # itertools.chain is in an extension module and may be unavailable
+        for it in its:
+            for x in it:
+                yield x
```
mgedmin: This looks very similar to [traceback._iter_chain() from Python 3.4](https://github.com/python/cpython/blob/3.4/Lib/traceback.py#L120-L138), but differs in some small details.   Where did it come from?

src/zope/testrunner/tb_format.py
```diff
@@ -20,9 +20,34 @@
 import zope.testrunner.feature
 
 
+try:
+    _iter_chain = traceback._iter_chain
+except AttributeError:
+    # Python 3.5
```
mgedmin: BTW Python 2.7 doesn't have `traceback._iter_chain` either, but I suppose there are no chained exceptions on Python 2, so that never came up.

src/zope/testrunner/tb_format.py
felixonmars: Looks like I looked up a somewhat outdated copy of that file, sorry for this. Link to the copy I used: http://svn.python.org/projects/python/branches/pep-0384/Lib/traceback.py

I'm going to update the PR using a more recent version (with a little adjustment to the "yield from" part, as that's an invalid syntax for Python 2)

mgedmin:
LGTM.

Do we need a new zope.testrunner release on PyPI to support Python 3.5 for other zope.* packages that use zope.testrunner to run their tests?

felixonmars:
I didn't run into problems with testrunner when running tests of other zope.* packages yet, but a new release would be safer :)

PR #36, opened 2016-03-24
Drop support for Python 2.6.

tseaver:


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py26,py27,pypy,py33,py34,py35,pypy3
+    py27,pypy,py33,py34,py35,pypy3
```
jamadden: Spurious [testenv:py26] below. Other than that LGTM.

PR #41, opened 2016-06-20
Fixed: Using the -j option to run tests in multiple processes

jimfulton:
Fixed: Using the -j option to run tests in multiple processes caused tests that used the multiprocessing package to hang (because the testrunner replaced sys.stdin with an unclosable object).


tseaver:
LGTM

PR #43, opened 2016-07-07
Make the 'subunit' support purely optional.

tseaver:
Applications which have been getting the dependencies via `zope.testrunner` should either add `zope.testrunner[subunit]` to their `install_requires`, or else depend directly on `python-subunit`.

The rationale for the change is that `python-subunit` is a heavy dependency, and that those who want the feature are likely to have it installed anyway.

/cc @multani, @rbtcollins, @menesis, @kedder, @mgedmin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,20 @@
 sudo: false
 language: python
-python:
-    - 2.7
-    - 3.3
-    - 3.4
-    - 3.5
-    - pypy
-    - pypy3
+matrix:
+  include:
+    - python: 3.5
+      env: TOXENV=py35
+env:
+    - TOXENV=py27
+    - TOXENV=py27-subunit
+    - TOXENV=pypy
+    - TOXENV=pypy-subunit
+    - TOXENV=py33
+    - TOXENV=py34
+    - TOXENV=pypy3
 install:
-    - pip install .
+    - pip install tox
 script:
-    - python setup.py -q test -q
+    - tox
 notifications:
     email: false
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -5,15 +5,6 @@ envlist =
 [testenv]
 deps =
     zope.testing
+    subunit: python-subunit
```
mgedmin: This bit seems underexplained.

.travis.yml
```diff
@@ -10,6 +10,6 @@ python:
 install:
     - pip install .
 script:
-    - python setup.py test -q
+    - python setup.py -q test -q
```
mgedmin: Are we not testing subunit support on Travis?

mgedmin:
Mostly LGTM, just worried that optional subunit support will bitrot if the subunit tests aren't enabled by default (by which I mean Travis and `tox` with no arguments).

BTW does subunit support Python 3 these days?

tox.ini
tseaver: [Factors](https://testrun.org/tox/latest/config.html#factors-and-factor-conditional-settings) are parts of environment names.  This line makes `py27-subunit`, `py34-subunit`, and `pypy-subunit` valid environments, with that dependency installed.

.travis.yml
tseaver: We could add a matrix which enabled the `py27-subunit` and `pypy-subunit` environments.

tseaver:
> BTW does subunit support Python 3 these days?

It does, although there seem to be bugs in it (doctest support, at least).

PR #48, opened 2016-12-27
Force Python 3.5 to work on Travis CI

mgedmin:
Reasons to accept this:
- it makes Python 3.5 pass on Travis CI
- it makes `tox --develop` pass locally
- it makes PyPy3 pass on Travis CI

Reasons to reject this:
- this is black magic workaround that papers around the problem instead of fixing it
- it conflates two fixes (py35 and pypy3) in one pull request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
     - python: 3.5
       env: TOXENV=py35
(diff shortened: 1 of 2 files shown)


mgedmin:
I'm going to merge this, since having the tests pass on master is important.

PR #55, opened 2017-11-10
Automatically enable the default warnings

### LINKED ISSUE #54: Running tests should enable DeprecationWarning, etc
[According to Nick Coghlan](https://mail.python.org/pipermail/python-dev/2017-November/150436.html), test runners are expected to override the default filters and enable the same warnings that [unittest test runners do](https://docs.python.org/3/library/unittest.html#unittest.TextTestRunner)

See also the pytest issue: https://github.com/pytest-dev/pytest/issues/2908

jamadden:
Fixes #54

The handling of the arguments and warnings is directly lifted from ``unittest``.

I introduced a new method, run_tests_with_warnings, that calls the pre-existing run_tests. In the (unlikely?) event of subclasses that override run_tests, this will ensure that they get the warnings by
default. It will also allow subclasses to change warning behaviour without changing test behaviour.

This started producing a bunch of ResourceWarnings under Python 3 for unclosed files and un-waited-for child process, and that broke a doctest, so I fixed that in spawn_layer_in_subprocess.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,13 +16,12 @@
 # When developing and releasing this package, please follow the documented
 # Zope Toolkit policies as described by this documentation.
 ##############################################################################
-version = '4.7.1.dev0'
-
 import os
 import sys
 from setuptools import setup
 from setuptools.command.test import test
 
+version = '4.8.0.dev0'
 
 INSTALL_REQUIRES = [
     'setuptools',
@@ -55,7 +54,7 @@ import zope.testing
 import zope.testrunner
 if __name__ == '__main__':
     zope.testrunner.run([
-        '--test-path', %r, '-c'
+        '--test-path', %r, '-c',
         ])
 """
 
@@ -77,12 +76,12 @@ class custom_test(test):
 
     def run_tests(self):
         import tempfile
-        fd, filename = tempfile.mkstemp(prefix='temprunner', text=True)
-        scriptfile = open(filename, 'w')
         script = CUSTOM_TEST_TEMPLATE % (
             sys.path, os.path.abspath(os.curdir), os.path.abspath('src'))
-        scriptfile.write(script)
-        scriptfile.close()
+        _fd, filename = tempfile.mkstemp(prefix='temprunner', text=True)
+        with open(filename, 'w') as scriptfile:
+            scriptfile.write(script)
+            scriptfile.close()
 
         import subprocess
         process = subprocess.Popen([sys.executable, filename])
@@ -96,8 +95,12 @@ class custom_test(test):
             sys.stderr.flush()
         sys.exit(rc)
 
+def read(*names):
+    with open(os.path.join(*names)) as f:
+        return f.read()
+
 chapters = '\n'.join([
-    open(os.path.join('src', 'zope', 'testrunner', 'tests', name)).read()
+    read('src', 'zope', 'testrunner', 'tests', name)
     for name in (
         'testrunner.txt',
         'testrunner-simple.txt',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -77,12 +76,12 @@ def run(self):
...
-        fd, filename = tempfile.mkstemp(prefix='temprunner', text=True)
-        scriptfile = open(filename, 'w')
         script = CUSTOM_TEST_TEMPLATE % (
             sys.path, os.path.abspath(os.curdir), os.path.abspath('src'))
-        scriptfile.write(script)
-        scriptfile.close()
+        _fd, filename = tempfile.mkstemp(prefix='temprunner', text=True)
+        with open(filename, 'w') as scriptfile:
+            scriptfile.write(script)
+            scriptfile.close()
```
mgedmin: You shouldn't need an explicit `close()` here.

src/zope/testrunner/runner.py
```diff
@@ -533,6 +573,14 @@ def spawn_layer_in_subprocess(result, script_parts, options, features,
...
     finally:
         result.done = True
+        if child is not None:
+            # Regardless of whether the process ran to completion, we
+            # must properly cleanup the process to avoid
+            # `ResourceWarning: subprocess XXX is still alive` and
+            # `ResourceWarning: unclosed file` for its stdout and
+            # stderr.
+            child.kill()
+            child.communicate()
```
mgedmin: I would mildly prefer `child.terminate()` and a little timeout loop calling `child.poll()` before resorting to `child.kill()`, on general principles.

src/zope/testrunner/tests/test_runner.py
```diff
@@ -173,10 +173,34 @@ class ZZ(K1, K2, K3): pass
...
+            warnings.simplefilter('default')
+            warnings.filterwarnings('module',
+                                    category=DeprecationWarning,
+                                    message=r'Please use assert\w+ instead.')
+            new_filters = [(f[0], f[2], 4) for f in warnings.filters]
+
+        # For some reason, catch_warnings doesn't fully reset things,
+        # and we wind up with some duplicate entries in new_filters
+        self.assertEqual(set(old_filters), set(new_filters))
+        warnings.warn("This should be visible by default", DeprecationWarning)
```
mgedmin: This is great while developing this feature, but shouldn't be left this way.  For one, it's not an automated test if it requires human verification.

Can we intercept the warning (maybe using `logging.captureWarnings(True)`) and assert that it was shown?

src/zope/testrunner/runner.py
jamadden: The Popen documentation [suggests using kill](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate).

setup.py
jamadden: D'oh!

src/zope/testrunner/runner.py
mgedmin: Yes, after a timeout.  I also want a timeout before `.kill()` ;)

Again, no specific reason, so I won't mind if you don't do that here.

src/zope/testrunner/tests/test_runner.py
jamadden: good idea.

src/zope/testrunner/runner.py
jamadden: At this point in the code, the child process *should* already be terminated. There's a loop above that's effectively polling for it to die (albeit without a timeout).

mgedmin: That sounds convincing!

jamadden:
Neat! Thanks!

Should we do a PyPI release? I have a feeling that this might break some doctests (but hopefully if it's ResourceWarnings, like it was here, fixing them will be worth it).

icemac:
Thank you for releasing this as 4.8.0 to PyPI.

PR #56, opened 2017-11-10
Mark wheels as universal and stop forcing sdists to be zips

jamadden:
.zip sdists are deprecated and on the way to unsupported. See
https://mail.python.org/pipermail/distutils-sig/2016-August/029384.html

Also adjust the badges in README to the standard and remove winbot since it is dead.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,8 +2,16 @@
 zope.testrunner
 ***************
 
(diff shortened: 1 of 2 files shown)


jamadden:
Thanks! I'll push 4.8 to PyPI now.

PR #58, opened 2017-11-10
Enable warnings during test discovery.

### LINKED ISSUE #57: Warnings should be enabled sooner to capture deprecated imports
I ran the test case for zope.app.apidoc with the new zope.testrunner 4.8 to check for warnings. I got some reported for zope.pluggableauth:
```
///zope/pluggableauth/authentication.py:23: DeprecationWarning: zope.site.next has moved to zope.component. Import of zope.site.next will become unsupported in Version 5.0
  from zope.site.next import queryNextUtility
///zope/pluggableauth/plugins/session.py:28: DeprecationWarning: zope.site.hooks has moved to zope.component.hooks. Import of zope.site.hooks will become unsupported in Version 5.0
  from zope.site import hooks
```

Happily I went to zope.pluggableauth and ran its own tests to check for more deprecations. Much to my surprise, there were *no* deprecations reported. Thinking about it, this makes sense: deprecations are currently only enabled when tests are being run, but tests have to be discovered before they can be run, which means importing them. And warnings aren't enabled at import time. (zope.deprecation is used for the warning and emits it at import.)

Here's the stack of where the import is being done:
```
  //site-packages/zope/testrunner/__init__.py(31)run()
-> warnings=warnings)
  //site-packages/zope/testrunner/__init__.py(51)run_internal()
-> runner.run()
  //site-packages/zope/testrunner/runner.py(170)run()
-> feature.global_setup()
  //site-packages/zope/testrunner/find.py(479)global_setup()
-> tests = find_tests(self.runner.options, self.runner.found_suites)
  //site-packages/zope/testrunner/find.py(168)find_tests()
-> for suite in found_suites:
  //site-packages/zope/testrunner/find.py(194)find_suites()
-> module = import_name(module_name)
  //site-packages/zope/testrunner/find.py(395)import_name()
```

jamadden:
Fixes #57

Refactor the warning enablement into its own (protected) method (since we need to use it during the already-public 'run' method) and go back to just using `run_tests` Subclasses can still override that method to change warning behaviour.

I did not add a specific test case for this. Because zope.testrunner runs its own tests, trying to test a feature that issued a warning would always succeed (because warnings are enabled when tests are run), without trying to do a bunch of messing with the filters or subprocesses.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ zope.testrunner Changelog
 4.8.1 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Enable ``DeprecationWarning`` earlier, when discovering test
+  modules. This lets warnings that are raised on import (such as those
+  produced by ``zope.deprecation.moved``) be reported. See `issue 57
+  <https://github.com/zopefoundation/zope.testrunner/issues/57>`_.
 
 
 4.8.0 (2017-11-10)
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -26,6 +26,8 @@ import traceback
 import unittest
 import warnings
 
+from contextlib import contextmanager
+
 from six import StringIO
 from zope.testrunner.find import import_name
 from zope.testrunner.find import name_from_layer, _layer_name_cache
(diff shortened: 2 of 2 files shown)


jamadden:
Thanks!

PR #59, opened 2017-11-13
Enable coverage measurement

jamadden:
- tox.ini environment
- coveralls.io on Travis
- Add badge
- Drop Python 3.3
- Some low-hanging fruit for improved coverage: remove code that was only executed on 2.6 or lower

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,12 @@
 [run]
 source = zope.testrunner
 parallel = true
+omit =
+    # These files are definitely imported and executed,
+    # but for some reason don't show up as such.
+    # See testrunner-knit
+    src/zope/testrunner/tests/testrunner-ex-pp-products/more/sampletests.py
+    src/zope/testrunner/tests/testrunner-ex-pp-products/sampletests.py
 
 [report]
 precision = 2
@@ -10,3 +16,5 @@ exclude_lines =
     raise NotImplementedError
     self.fail
     raise AssertionError
+    if is_jython:
+    if sys.platform == 'win32':
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,12 @@
+[run]
+source = zope.testrunner
+parallel = true
+
+[report]
+precision = 2
+exclude_lines =
+    pragma: no cover
+    if __name__ == '__main__':
+    raise NotImplementedError
+    self.fail
+    raise AssertionError
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ develop-eggs/
 tags
 .tox
 parts/
+.coverage
+htmlcov/
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ tags
 .tox
 parts/
 .coverage
+.coverage.*
 htmlcov/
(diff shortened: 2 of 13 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,9 @@ zope.testrunner Changelog
 4.8.2 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Drop support for Python 3.3.
```
mgedmin: I wonder if that makes it a 4.9.0.

src/zope/testrunner/profiling.py
```diff
@@ -26,7 +25,7 @@
 try:
     import cProfile
     import pstats
-except ImportError:
+except ImportError: # pragma: no cover (Where could this fail?)
```
mgedmin: A long time ago Linux distros used to split off `profile` and `pstats` modules into a [separate python-profile distro package](https://bugs.launchpad.net/ubuntu/+source/python-defaults/+bug/123755) for [licencing reasons](https://bugs.python.org/issue12417).  This was resolved in 2011.

mgedmin: (Also, `cProfile` was not available in Python 2.4, and zope.testrunner *is* that old.)

.coveragerc
```diff
@@ -0,0 +1,20 @@
+[run]
+source = zope.testrunner
+parallel = true
+omit =
+    # These files are definitely imported and executed,
+    # but for some reason don't show up as such.
+    # See testrunner-knit
+    src/zope/testrunner/tests/testrunner-ex-pp-products/more/sampletests.py
+    src/zope/testrunner/tests/testrunner-ex-pp-products/sampletests.py
```
mgedmin: Looking at `testrunner-knit.txt`, it may be that coverage.py doesn't handle the special runtime Python package's `__path__` variable manipulation magic that this test performs?

I don't quite understand what that's supposed to be used for -- pre-PEP-420 namespace-packages?

.gitignore
```diff
@@ -9,3 +9,5 @@ develop-eggs/
 tags
 .tox
 parts/
+.coverage
```
mgedmin: Given `parallel = true`, maybe also ignore `.coverage.*`?  OTOH those are deleted when you (or tox) run `coverage combine`, so maybe we don't care.

src/zope/testrunner/profiling.py
mgedmin: Anyway, I'm :+1: for not handling `ImportError`s from cProfile and pstats in 2017.

CHANGES.rst
jamadden: tseaver would say so, yes. I'll bump.

.coveragerc
jamadden: I've speculated [about similar uses before](https://github.com/zopefoundation/ZConfig/issues/28#issuecomment-287318917) in other projects. It's not clear to me either.

.gitignore
jamadden: Seems reasonable enough. If a subprocess crashes or fails then tox doesn't run the next command, right? So failing tests would leave these lying around.

jamadden:
Thank you!

Issue #35, opened 2016-02-09
Test failure on Python3.4


domenkozar:
When upgrading wheel/setuptools/pip, following test failure happens:

```
File "/nix/store/0imhpb40v7k9kvgx0adg6hihw8c6aw7l-python3.4-zope.testrunner-4.4.10/lib/python3.4/site-packages/zope/testrunner/tests/testrunner-test-selection.txt", line 236, in testrunner-test-selection.txt
Failed example:
    testrunner.run_internal(defaults)
Expected:
    Running tests at level 1
    Running zope.testrunner.layer.UnitTests tests:
      Set up zope.testrunner.layer.UnitTests in N.NNN seconds.
      Running:
     test_y0 (sample1.sampletests.test1.TestA)
     test_x0 (sample1.sampletests.test1.TestB)
     test_x0 (sample1.sampletests.test1)
     test_y0 (sample1.sampletests.test1)
     test_y0 (sample1.sampletests.test_one.TestA)
     test_x0 (sample1.sampletests.test_one.TestB)
     test_x0 (sample1.sampletests.test_one)
     test_y0 (sample1.sampletests.test_one)
      Ran 8 tests with 0 failures, 0 errors and 0 skipped in N.NNN seconds.
    Tearing down left over layers:
      Tear down zope.testrunner.layer.UnitTests in N.NNN seconds.
    False
Got:
    Running tests at level 1
    Running zope.testrunner.layer.UnitTests tests:
      Set up zope.testrunner.layer.UnitTests in N.NNN seconds.
      Running:
     test_y0 (sample1.sampletests.test1.TestA)
     test_x0 (sample1.sampletests.test1.TestB)
     test_x0 (sample1.sampletests.test1)
     test_y0 (sample1.sampletests.test1)
     testrunner-ex/sample1/sampletests/../../sampletests.txt
     test_y0 (sample1.sampletests.test_one.TestA)
     test_x0 (sample1.sampletests.test_one.TestB)
     test_x0 (sample1.sampletests.test_one)
     test_y0 (sample1.sampletests.test_one)
     testrunner-ex/sample1/sampletests/../../sampletests.txt
      Ran 10 tests with 0 failures, 0 errors and 0 skipped in N.NNN seconds.
    Tearing down left over layers:
      Tear down zope.testrunner.layer.UnitTests in N.NNN seconds.
    False

  Ran 53 tests with 1 failures, 0 errors, 0 skipped in 20.575 seconds.
```

[...]

PR #66, opened 2017-11-23
Workaround for issue 35

mgedmin:
See https://github.com/zopefoundation/zope.testrunner/issues/35 for all
the gory details.


orivej:
Thank you! This will prevent test failures in Nixpkgs.

jamadden:
I am currently traveling. It is likely to be some days before I have a chance to properly review anything.

mgedmin:
I don't think this is particularly urgent.

Have a good Thanksgiving, or however you say it in your distant country!

PR #67, opened 2017-12-24
Add option to check for duplicate test IDs

cjwatson:
Add a --require-unique flag to the testrunner.  When passed, this flag
causes the testrunner to check every test ID it loads to see if it's
seen it before.  If it has, it raises an error and exits.

This was originally by Graham Binns in
https://code.launchpad.net/~gmb/zope.testing/check-for-unique-tests/+merge/110262,
in response to https://bugs.launchpad.net/launchpad/+bug/682771, but
doesn't appear to have been sent upstream.  I ported it across to
zope.testrunner and generally brought it up to date.

### DIFF
--- a/src/zope/testrunner/find.py
+++ b/src/zope/testrunner/find.py
@@ -28,6 +28,10 @@ from zope.testrunner.filter import build_filtering_func
 identifier = re.compile(r'[_a-zA-Z]\w*$').match
 
 
+class DuplicateTestIDError(Exception):
+    """Raised whenever a test ID is encountered twice during loading."""
+
+
 class StartUpFailure(unittest.TestCase):
     """Empty test case added to the test suite to indicate import failures.
 
@@ -159,6 +163,7 @@ def find_tests(options, found_suites=None):
     """
     remove_stale_bytecode(options)
     suites = {}
+    dupe_ids = set()
 
     test_accept = build_filtering_func(options.test)
     module_accept = build_filtering_func(options.module)
@@ -167,11 +172,21 @@ def find_tests(options, found_suites=None):
         found_suites = find_suites(options, accept=module_accept)
     for suite in found_suites:
         for test, layer_name in tests_from_suite(suite, options,
-                                                 accept=test_accept):
+                                                 accept=test_accept,
+                                                 duplicated_test_ids=dupe_ids):
+            if dupe_ids:
+                # If there are any duplicated test IDs, we stop trying to
+                # load tests; we'll raise an error later on with all the
+                # duplicates in it.
+                continue
             suite = suites.get(layer_name)
             if not suite:
                 suite = suites[layer_name] = unittest.TestSuite()
             suite.addTest(test)
+    if dupe_ids:
+        message_lines = ['Duplicate test IDs found:'] + sorted(dupe_ids)
+        message = '\n  '.join(message_lines)
+        raise DuplicateTestIDError(message)
     return suites
 
 
@@ -398,7 +413,8 @@ def import_name(name):
 
 def tests_from_suite(suite, options, dlevel=1,
                      dlayer=zope.testrunner.layer.UnitTests,
-                     accept=None):
+                     accept=None,
+                     seen_test_ids=None, duplicated_test_ids=None):
     """Returns a sequence of (test, layer_name)
 
     The tree of suites is recursively visited, with the most specific
@@ -410,6 +426,14 @@ def tests_from_suite(suite, options, dlevel=1,
     accept is a function, returning boolean for given test name (see also
     build_filtering_func()).
     """
+    # We use this to track the test IDs that have been registered.
+    # tests_from_suite will complain if it encounters the same test ID
+    # twice.
+    if seen_test_ids is None:
+        seen_test_ids = set()
+    if duplicated_test_ids is None:
+        duplicated_test_ids = set()
+
     level = getattr(suite, 'level', dlevel)
     layer = getattr(suite, 'layer', dlayer)
     if not isinstance(layer, six.string_types):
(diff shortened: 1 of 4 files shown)


src/zope/testrunner/find.py
```diff
@@ -418,11 +442,19 @@ def tests_from_suite(suite, options, dlevel=1,
...
     else:
+        if options.require_unique_ids:
+            suite_id = suite.id()
+            if suite_id in seen_test_ids:
+                duplicated_test_ids.add(suite_id)
+            else:
+                seen_test_ids.add(suite_id)
         if level <= options.at_level:
             if accept is None or accept(str(suite)):
                 yield (suite, layer)
```
mgedmin: Note that you're enforcing unique `suite.id()`, but the filtering checks `str(suite)`.  These are different:

- for regular unittest.TestCase tests `id()` returns `package.module.ClassName.test_method`, while `str()` returns `test_method (package.module.ClassName)`
- for doctest.DocFileSuite tests, `id()` returns `filename_txt`, while `str()` returns `/absolute/path/to/the/filename.txt`.

(Incidenally, this is why the original bug report doesn't apply to the test runner: we can always distinguish different distribution.txt doctest files if they're in different directories).

src/zope/testrunner/options.py
```diff
@@ -647,6 +655,15 @@ def apply_auto_progress(args):
...
         return options
 
+    if options.module and options.require_unique_ids:
+        # We warn if --module and --require-unique are specified at the same
+        # time, though we don't exit.
+        print("""\
+        You specified a module along with --require-unique;
+        --require-unique will not try to enforce test ID uniqueness when
+        working with a specific module.
+        """)
```
mgedmin: The same applies to `--package` filtering, doesn't it?

mgedmin:
Note that zope.testrunner's use of `str(test)` for identifying tests has some downsides:

- it's harder to match test classes (`-t 'test_method \(.*.TestClass\)'`)
- doctest file parent directories might accidentally match when you don't want them to match (see bug #35)

However using `test.id()` would have the problem of making it nearly impossible to figure out where doctest files live, and I suppose seeing the test name at the beginning of the string may make it easier to understand which test has failed.

I've a [vague wish/plan](https://github.com/zopefoundation/zope.testrunner/issues/35#issuecomment-346583668) of migrating zope.testrunner to a custom test ID function that at least strips off
parent directory names for doctest files.  The first step would be to refactor the existing codebase so instead of using `str(test)` it would call `test_name(test)` everywhere.  A quick git grep shows

```
src/zope/testrunner/find.py:427:            if accept is None or accept(str(suite)):
src/zope/testrunner/formatter.py:77:        s = str(test)
src/zope/testrunner/formatter.py:290:            s = str(test)
src/zope/testrunner/process.py:46:            print(' '.join(str(test).strip().split('\n')), file=self.original_stderr)
src/zope/testrunner/process.py:48:            print(' '.join(str(test).strip().split('\n')), file=self.original_stderr)
```

Hm.  Those last lines make me suspect I mixed something up in the description above -- surely str(test) shouldn't have embedded spaces, that would break the subprocess protocol!

src/zope/testrunner/find.py
```diff
@@ -418,11 +442,19 @@ def tests_from_suite(suite, options, dlevel=1,
...
+                                      seen_test_ids=seen_test_ids,
+                                      duplicated_test_ids=duplicated_test_ids):
                 yield r
     elif isinstance(suite, StartUpFailure):
         yield (suite, None)
     else:
+        if options.require_unique_ids:
+            suite_id = suite.id()
+            if suite_id in seen_test_ids:
+                duplicated_test_ids.add(suite_id)
```
mgedmin: (Augh, GitHub cuts off the one line I hoped would be visible above my comment.)

TL;DR: I think you should change this to `suite_id = str(suite)`, and it's all good!

cjwatson:
Yeah, in Launchpad we override `TestCase.__str__` to return `test.id()`, and we use the equivalent of `os.path.relpath(test.id(), tree_root)` for doctests (I *think* the latter is a customisation we apply somewhere, although I haven't yet figured out exactly where so I could be wrong).  That seems to work fairly well: in particular it means that you can use things like `-t lp.bugs` to run all the tests in a given package regardless of whether they're unit tests or doctests, relying on regex `.` matching `/`.

Anyway, I've made the change you suggest.

mgedmin:
The Appveyor failures are caused by Windows path separators being `\\` instead of `/`.  Shouldn't be too hard to fix.

src/zope/testrunner/tests/test_find.py
```diff
@@ -0,0 +1,79 @@
...
+            find.DuplicateTestIDError,
+            find.find_tests, UniquenessOptions(), [self.test_suites])
+
+    def test_DuplicateTestIDError_message_contains_all_test_ids(self):
+        # The error raised by find_tests when duplicates are encountered
+        # contains all the duplicate test IDs it found.
+        with self.assertRaises(find.DuplicateTestIDError) as e:
+            find.find_tests(UniquenessOptions(), [self.test_suites])
+        self.assertIn('testrunner-ex/sampletests.rst', str(e.exception))
+        self.assertIn('testrunner-ex/sampletestsl.rst', str(e.exception))
```
mgedmin: This is the failing test on Windows.  `os.path.join('testrunner-ex', 'sampletests.rst')` should make it work.

mgedmin:
Have you signed the [Zope Foundation Committer Agreement](https://docs.zope.org/developer/becoming-a-committer.html)?  I'm not allowed to merge code from people who haven't signed it.  (And once you sign it, you'll get the committer bit that allows you to hit the nice big green Merge button yourself.)

cjwatson:
Thanks, I've pushed that Windows fix.

I haven't yet signed the committer agreement, though certainly can.  I'm not sure my home printer/scanner works at the moment, so there might be a short delay until I can make a trip to the library ...

PR #69, opened 2017-12-31
Refactor subprocess result output to emit bytes

cjwatson:
This fixes the underlying problems that led to test failures on Python 3
in #23, paving the way for the reintroduction of subunit support.  It
also generally makes more logical sense to emit bytes here, as can be
seen from the way that we no longer need to decode in the Python 3 case
only to have the stream encode again.

Doctests still need to work with text, but we can handle that by
monkey-patching a replacement for `doctest._SpoofOut` on Python 3 that
makes an underlying `io.BytesIO` available.  (This effectively takes the
run-time special cases for Python 3 and isolates them in the doctest
support code, which seems like an improvement.)

`_get_output_buffer` is a simplified version of `subunit._unwrap_text`,
incorporated here to allow consistent behaviour without a hard
dependency on subunit.

### DIFF
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -15,10 +15,12 @@
 """
 from __future__ import print_function
 
-import subprocess
 import errno
 import gc
+import io
+import os
 import re
+import subprocess
 import sys
 import threading
 import time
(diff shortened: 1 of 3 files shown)


cjwatson:
I finally managed to get the line-ending stuff to work; @mgedmin, can you recheck?

mgedmin:
It looks like this PR introduced regressions: https://github.com/zopefoundation/ZEO/pull/117#issuecomment-429639082

PR #70, opened 2018-01-10
Reintroduce subunit support

cjwatson:
This reintroduces the support for subunit that was dropped in 4.7.0, and
adds support for the subunit v2 protocol as suggested by @rbtcollins in
#23.  The runner now makes bytes/text distinctions that match those
expected by subunit, and the subunit options are now tested on all
supported Python versions.

There are still reasons why users might prefer to use subunit v1: for
instance, they might be processing the output using a version of subunit
that's too old to support the v2 protocol, or they might rely on the
distinction between failures and errors
(https://bugs.launchpad.net/subunit/+bug/1740158).  I've therefore
preserved the `--subunit` option for v1 and added a new `--subunit-v2`
option.

### DIFF
--- a/src/zope/testrunner/tests/testrunner-subunit-v2.rst
+++ b/src/zope/testrunner/tests/testrunner-subunit-v2.rst
@@ -9,7 +9,7 @@ protocol.
 First we need to make a temporary copy of the entire testing directory:
 
     >>> import os.path, sys, tempfile, shutil
-    >>> tmpdir = tempfile.mkdtemp()
+    >>> tmpdir = tempfile.mkdtemp(prefix='zope.testrunner-test-')
     >>> directory_with_tests = os.path.join(tmpdir, 'testrunner-ex')
     >>> source = os.path.join(this_directory, 'testrunner-ex')
     >>> n = len(source) + 1
@@ -188,7 +188,7 @@ themselves is often a critical part of the development process. Thus, it's
 good to be able to profile a test run.
 
     >>> import tempfile
-    >>> tempdir = tempfile.mkdtemp(prefix='zope.testrunner-')
+    >>> tempdir = tempfile.mkdtemp(prefix='zope.testrunner-test-')
 
     >>> sys.argv = [
     ...     'test', '--layer=122', '--profile=cProfile', '--subunit-v2',
--- a/src/zope/testrunner/tests/testrunner-subunit-v2.rst
+++ b/src/zope/testrunner/tests/testrunner-subunit-v2.rst
@@ -64,19 +64,22 @@ For easier doctesting, we use a helper that summarizes the output.
 
     >>> def subunit_summarize(func, *args, **kwargs):
     ...     orig_stdout = sys.stdout
-    ...     if sys.version_info[0] >= 3:
-    ...         import io
-    ...         sys.stdout = io.TextIOWrapper(io.BytesIO(), encoding='utf-8')
-    ...     else:
-    ...         import StringIO
-    ...         sys.stdout = StringIO.StringIO()
-    ...     ret = func(*args, **kwargs)
-    ...     sys.stdout.flush()
-    ...     buf = sys.stdout
-    ...     if sys.version_info[0] >= 3:
-    ...         buf = buf.detach()
-    ...     buf.seek(0)
-    ...     sys.stdout = orig_stdout
+    ...     try:
+    ...         if sys.version_info[0] >= 3:
+    ...             import io
+    ...             sys.stdout = io.TextIOWrapper(
+    ...                 io.BytesIO(), encoding='utf-8')
+    ...         else:
+    ...             import StringIO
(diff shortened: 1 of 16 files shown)


src/zope/testrunner/tests/testrunner-subunit.rst
```diff
@@ -0,0 +1,721 @@
...
+    $ pip install zope.testrunner[subunit]
+
+or directly:
+
+    $ pip install python-subunit
+
+First we need to make a temporary copy of the entire testing directory:
+
+    >>> import os.path, sys, tempfile, shutil
+    >>> tmpdir = tempfile.mkdtemp()
```
mgedmin: Please `prefix='test-zope.testrunner-'`.

(I'm probably wildly inconsistent with my suggested prefixes.  The main thing I want is that it mention zope.testrunner, and that it was created from test code, to help narrow down the source of mysterious leftover temp files.)

mgedmin: (The experience of running a Zope 3 buildbot on a machine with a small-by-modern-standards /tmp partition left scars.)

src/zope/testrunner/tests/testrunner-subunit.rst
```diff
@@ -0,0 +1,721 @@
...
+    >>> directory_with_tests = os.path.join(tmpdir, 'testrunner-ex')
+    >>> source = os.path.join(this_directory, 'testrunner-ex')
+    >>> n = len(source) + 1
+    >>> for root, dirs, files in os.walk(source):
+    ...     dirs[:] = [d for d in dirs if d != ".svn"] # prune cruft
```
mgedmin: I don't think we need to prune `.svn` directories any more.

src/zope/testrunner/tests/testrunner-subunit.rst
```diff
@@ -0,0 +1,721 @@
...
+    >>> from zope import testrunner
+
+
+Basic output
+------------
+
+Subunit output is line-based, with a 'test:' line for the start of each test
+and a 'successful:' line for each successful test.
+
+Zope layer set up and tear down events are represented as tests tagged with
```
mgedmin: I think it's "setup and teardown events" when "set up" and "tear down" are not verbs, but I'm not a native English speaker.

src/zope/testrunner/tests/testrunner-subunit.rst
```diff
@@ -0,0 +1,721 @@
...
+is an error containing the traceback. These "tests" are tagged with
+zope:import_error.
+
+Let's create a module with some bad syntax:
+
+    >>> badsyntax_path = os.path.join(directory_with_tests,
+    ...                               "sample2", "sampletests_i.py")
+    >>> f = open(badsyntax_path, "w")
+    >>> print("importx unittest", file=f)  # syntax error
+    >>> f.close()
```
mgedmin: Same concerns about writing to the source tree apply here too.

src/zope/testrunner/tests/testrunner-subunit-v2.rst
```diff
@@ -0,0 +1,654 @@
...
+      tags=(zope:layer:sample3.sampletests_ntd.Layer)
+    id=sample3.sampletests_ntd.Layer:tearDown status=inprogress !runnable
+    id=sample3.sampletests_ntd.Layer:tearDown !runnable
+    reason (text/plain...)
+    tearDown not supported
+    id=sample3.sampletests_ntd.Layer:tearDown status=skip tags=(zope:layer)
+      !runnable
+    True
+
+Note that debugging doesn't work when running tests in a subprocess:
```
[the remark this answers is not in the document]
cjwatson: I agree that this is unlikely to go well.  I'm slightly reluctant to actively forbid it, though, since there might be some situation I haven't thought of in which it's a useful last resort (maybe a test breaks only in `--subunit-v2` mode for some weird reason and you need to debug it despite the likely terminal breakage).

Do you think there's a reason to actively go to the effort of forbidding this particular case, which seems like just a special case of "using `--subunit-v2` when stdout is a terminal is probably a bad idea"?

cjwatson:
I think I've fixed all the things you pointed out, with the exception of one inline question; please re-review?

src/zope/testrunner/tests/testrunner-subunit-v2.rst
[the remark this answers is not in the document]
mgedmin: No, I was just wondering.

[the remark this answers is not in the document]
rbtcollins: You can actually use the stdout capturing to tunnel debug sessions across machines - in principle all through the testr stack as well, though there is some missing glue to make that work completely (I need a terminal multiplexer :().

Put another way: you can drop a subunit filter onto the terminal if you're running tests in subunit mode and want to debug without seeing the protocol itself.

PR #71, opened 2018-01-19
Improve output when layer setUp or tearDown fails

cjwatson:
In particular, the subunit output is now much more precise in this case.

This is based on work by Graham Binns, Gary Poster, and Francesco
Banconi in Launchpad's zope.testing fork.

(More in the ongoing saga of "unearth changes made before my time and try to sort them out into something vaguely upstreamable ...")

### DIFF
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -64,7 +64,6 @@ class UnexpectedSuccess(Exception):
     pass
 
 
-EXPLOSIVE_ERRORS = (MemoryError, KeyboardInterrupt, SystemExit)
 PYREFCOUNT_PATTERN = re.compile(r'\[[0-9]+ refs\]')
 
 is_jython = sys.platform.startswith('java')
@@ -463,7 +462,7 @@ def run_layer(options, layer_name, layer, tests, setup_layers,
         setup_layer(options, layer, setup_layers)
     except zope.testrunner.interfaces.EndRun:
         raise
-    except EXPLOSIVE_ERRORS:
+    except MemoryError:
         raise
     except Exception:
         handle_layer_failure(SetUpLayerFailure(layer), output, errors)
@@ -793,7 +792,7 @@ def tear_down_unneeded(options, needed, setup_layers, errors, optional=False):
                 output.tear_down_not_supported()
                 if not optional:
                     raise CanNotTearDown(l)
-            except EXPLOSIVE_ERRORS:
+            except MemoryError:
                 raise
             except Exception:
                 handle_layer_failure(TearDownLayerFailure(l), output, errors)
@@ -821,7 +820,7 @@ def setup_layer(options, layer, setup_layers):
         if hasattr(layer, 'setUp'):
             try:
                 layer.setUp()
-            except EXPLOSIVE_ERRORS:
+            except MemoryError:
                 raise
             except Exception:
                 if options.post_mortem:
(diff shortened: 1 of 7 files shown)


src/zope/testrunner/runner.py
```diff
@@ -317,11 +318,22 @@ def run_tests(self):
...
         self.failed = bool(self.import_errors or self.failures or self.errors)
 
 
+def handle_layer_failure(failure_type, output, errors):
+    if hasattr(output, 'layer_failure'):
+        output.layer_failure(failure_type.subunit_label, sys.exc_info())
+    else:
+        f = StringIO()
+        traceback.print_exc(file=f)
+        output.error(f.getvalue())
```
mgedmin: Why the StringIO() dance rather than `traceback.format_exc()`?  (I've a feeling there's a good reason.)

src/zope/testrunner/runner.py
```diff
@@ -451,11 +463,10 @@ def run_layer(options, layer_name, layer, tests, setup_layers,
         setup_layer(options, layer, setup_layers)
     except zope.testrunner.interfaces.EndRun:
         raise
+    except EXPLOSIVE_ERRORS:
+        raise
     except Exception:
```
mgedmin: KeyboardInterrupt and SystemExit all inherit BaseException rather than Exception, and are not caught by the `except Exception` catch-all below.  I'm not sure how I feel about the redundant catch-and-release for them.

(I thought MemoryError was the same, but then I checked.)

src/zope/testrunner/runner.py
mgedmin: Ah, you moved preexisting code here, so I may be asking the wrong person.

src/zope/testrunner/tests/testrunner-subunit-v2.rst
```diff
@@ -291,8 +291,8 @@ https://bugs.launchpad.net/subunit/+bug/1740158.)
     True
 
 
-Layers that can't be torn down
-------------------------------
+Layer failures
+--------------
```
mgedmin: (Reviewing code while tired is not a good idea: I've feelings, but no articulatory power to express them well.)

To me it's not a *failure* when a layer's tearDown raises NotImplementedError -- it's an explicitly supported way of stating "tear down is not supported by this layer."  (It's a rather strange way -- I'd've preferred the layer to set tearDown=None, or something, so this fact could be discovered in advance and the test runner could punish the bad layer by running it in a subprocess, instead of running all the other innocent layers, but I digress.)

What I'm saying, I'd prefer the section about layer failures to be separate from layers that don't support teardown.

icemac:
What is the current status of this PR? What needed to be done before it can be merged?

icemac:
@cjwatson Ping.

src/zope/testrunner/runner.py
cjwatson: Ack - I suspect this code may have been very old, maybe even from before the BaseException rework.  Simplified.

src/zope/testrunner/tests/testrunner-subunit-v2.rst
cjwatson: OK, I've moved this into its own section.

cjwatson:
Sorry for the long delay.  I think it's basically ready now, but #95 needs to be reviewed first and then I'll need to rebase this branch again to get CI to pass.

mgedmin:
Thank you!

PR #73, opened 2018-09-20
Avoid aborting test run if a traceback can't be formatted

icemac:
@davisagli Do you think this commit should be included in the master branch?

### DIFF
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -453,8 +453,12 @@ def run_layer(options, layer_name, layer, tests, setup_layers,
         raise
     except Exception:
         f = StringIO()
-        traceback.print_exc(file=f)
-        output.error(f.getvalue())
+        try:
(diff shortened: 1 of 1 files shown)


src/zope/testrunner/runner.py
```diff
@@ -451,8 +451,12 @@ def run_layer(options, layer_name, layer, tests, setup_layers,
...
     except Exception:
         f = StringIO()
-        traceback.print_exc(file=f)
-        output.error(f.getvalue())
+        try:
+            traceback.print_exc(file=f)
+            output.error(f.getvalue())
+        except Exception:
+            output.error(
+                '*** An exception occurred, but could not be formatted ***')
```
mgedmin: It would be useful to see (a) the name and str() of the original exception, and (b) the name and str() of the exception that occurred while trying to format the original exception.

icemac:
I am not going to pursue this PR any further, so I am closing it.

PR #76, opened 2018-10-03
Added support for Python 3.7

vernans:


jamadden:
I don't see any 3.7 builds [on Travis](
https://travis-ci.org/zopefoundation/zope.testrunner/builds/436567123?utm_source=github_status&utm_medium=notification). I suspect that the `env` block and `matrix` block are conflicting.

.travis.yml
```diff
@@ -7,6 +7,16 @@ python:
   - 3.6
   - pypy
   - pypy3
+include:
+    matrix:
```
mgedmin: It's supposed to be the other way around, i.e.
```
matrix:
    include:
```

CHANGES.rst
```diff
@@ -28,6 +28,8 @@
 - Reintroduce optional support for ``subunit``, now with support for both
   version 1 and version 2 of its protocol.
 
+- Added support for Python 3.7.
```
mgedmin: Maybe move this next to the 'Dropped support for Python 3.3' line?  Or, alternatively, move that line down.

(I think it makes sense to keep supported Python version notes next to each other in the changelog.)

vernans: Yes, you're right. I'll do that from now on. 
Thank you for fixing it!

PR #82, opened 2018-11-21
Prefer the Thread.is_alive spelling

### LINKED ISSUE #81: "zope.testrunner" can crash due to "https://bugs.python.org/issue35283"
"zope.testrunner.runner.TestResult.stopTest" uses the thread method "isAlive". Due to "https://bugs.python.org/issue35283", this causes the runner to die with an "AssertionError" when it happens to be called on a "_DummyThread".

"_DummyThread" instances are internally created by Python's "threading" module. When I have understood the explanation in the "threading" comments correctly,  this happens when a thread started with "_thread.start_new_thread" calls (maybe indirectly) "threading.current_thread".

I recommend to consistently replace the use of "threading.Thread.isAlive" by the PEP 8 conformant "threading.Thread.is_alive" to work around the above Python bug.

mgedmin:
Not only are old camelCase names deprecated (and undocumented on Python
3), they're also broken (on Python 3).

Fixes #81.

### DIFF
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -705,7 +705,7 @@ def resume_tests(script_parts, options, features, layers, failures, errors,
             running_threads.append(thread)
 
         for index, thread in reversed(list(enumerate(running_threads))):
-            if not thread.isAlive():
+            if not thread.is_alive():
                 del running_threads[index]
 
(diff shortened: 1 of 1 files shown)


mgedmin:
Whoops, accidentally pushed the changelog update to master.

PR #83, opened 2018-11-23
Encode layer output if needed as UTF-8 before writing it to a buffer …

### LINKED ISSUE #80: TypeError when running tests in parallel
See https://github.com/zopefoundation/ZEO/pull/117#issuecomment-429639082

It seems to be caused by zopefoundation/zope.testrunner#69.

jamadden:
…that explicitly (only) accepts bytes.

Update the test case to make this explicit. There are three test failures without doing the encoding.

Fixes #80

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@
 4.9.2 (unreleased)
 ==================
 
-- Nothing changed yet.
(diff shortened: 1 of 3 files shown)


jamadden:
Thanks! I'll push a release to PyPI.

PR #89, opened 2019-02-12
Fix test failures on Python 3.8.0a1

mgedmin:
Prompted by https://github.com/zopefoundation/zopetoolkit/pull/32.

Tested with `tox -e py38`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: python
 python:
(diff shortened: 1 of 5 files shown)


mgedmin:
I do not like to claim Python 3.8 compatibility at this point, because Python 3.8 does not exist yet, and we cannot know 3.8 final won't uncover some other issues.

PR #90, opened 2019-03-15
Drop support for Python 3.4.

icemac:
Python 3.4 end of life is at 2019-03-16, see https://devguide.python.org/#status-of-python-branches

Note to self: After this PR is merge a release could be cut to help fixing zopefoundation/zopetoolkit#32.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{27,py,34,35,36,37}{,-subunit},coverage,docs
+    py{27,py,35,36,37,38}{,-subunit},coverage,docs
 
 [testenv]
 extras =
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
   - 2.7
-  - 3.4
   - 3.5
   - 3.6
   - pypy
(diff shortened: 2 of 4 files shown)


mgedmin:
AFAICT the March 16 date is only tentative and has passed without an EOL announcement or the [promised 3.4.10 final release](https://mail.python.org/pipermail/python-announce-list/2019-March/012186.html).  So, technically 3.4 is not EOL yet -- but I expect it will be within a couple of days.

Anyway, let's not split hairs, LGTM.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{27,py,34,35,36,37}{,-subunit},coverage,docs
+    py{27,py,35,36,37,38}{,-subunit},coverage,docs
```
mgedmin: Was adding py38 here intentional?

icemac:
@mgedmin Python 3.4.10 got announced just yesterday: [Python 3.4.10 is now available](https://pythoninsider.blogspot.com/2019/03/python-3410-is-now-available.html).

tox.ini
icemac: Yes it was because it is already tested via TravisCI and I'd like to keep tox.ini and .travis.yml in sync.

PR #97, opened 2019-10-15
Add test buffering, enabled by default for subunit

cjwatson:
Python 2.7 added a --buffer option to the unittest runner with the
following description:

  "The standard output and standard error streams are buffered during
  the test run.  Output during a passing test is discarded.  Output is
  echoed normally on test fail or error and is added to the failure
  messages."

Add a matching option to zope.testrunner.

This is important for subunit output: tests that output extra things to
stdout may corrupt the subunit stream (especially with v1).  This option
is therefore always enabled for --subunit and --subunit-v2.

This is based loosely on work by Graham Binns, Benji York, and Brad
Crittenden in Launchpad's zope.testing fork; but I took a different
approach that's much closer to how modern unittest works, and I haven't
yet bothered trying to buffer output from layers (on the assumption that
there will be many fewer layers than tests, making it much easier to fix
them individually to be quiet).

### DIFF
--- a/src/zope/testrunner/formatter.py
+++ b/src/zope/testrunner/formatter.py
@@ -364,15 +364,17 @@ class OutputFormatter(object):
     def print_std_streams(self, stdout, stderr):
         """Emit contents of buffered standard streams."""
         if stdout:
-            sys.stdout.write("\nStdout:\n")
+            sys.stdout.write("Stdout:\n")
             sys.stdout.write(stdout)
             if not stdout.endswith("\n"):
                 sys.stdout.write("\n")
+            sys.stdout.write("\n")
         if stderr:
-            sys.stderr.write("\nStderr:\n")
+            sys.stderr.write("Stderr:\n")
             sys.stderr.write(stderr)
             if not stderr.endswith("\n"):
                 sys.stderr.write("\n")
+            sys.stderr.write("\n")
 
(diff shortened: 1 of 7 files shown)


cjwatson:
CCing @rbtcollins into this PR in case he recalls any concerns with enabling stdout/stderr buffering by default for subunit.  (I'm attempting to upstream stuff that was done well before my time, so it's possible that I'm missing some context.)

cjwatson:
I found a problem when doing a bit more testing of this: it eats test names printed by `-vv`.  To avoid this I think I'll need to either sacrifice buffering around `testSetUp` or at least be more careful about it somehow.  I'll also look into making sure that the test suite catches this.

cjwatson:
OK, I've fixed the above problem.  @mgedmin, are you still OK with this approach?

cjwatson:
No, output from `setUp` and `tearDown` will be buffered, since those are called between `startTest` and whichever result method restores the streams.  It's just output from the layer's `testSetUp` and `testTearDown` methods that won't be buffered.  (The test protocol really is labyrinthine.)

I have indeed considered having `TestResult` (or perhaps `OutputFormatter`) store its own references and write directly to those.  I think it would probably be easier to follow and permit further improvements.  However, it feels like something to do in a separate PR; I wouldn't be surprised if it were quite fiddly to get working exactly right.

PR #99, opened 2019-10-17
Fix failure counting across layers

cjwatson:
When tests in multiple layers are being run, zope.testrunner outputs a
per-layer summary after each layer, and an overall summary at the end.
The addition of support for unexpected successes in 4.0.1 caused the
per-layer summary to report incorrect failure counts: rather than
reporting the number of failures in the current layer (as for errors and
skips), it reported the cumulative number of failures so far in the
whole test run.  This was confusing, since it could result in per-layer
summaries that appeared to report failures in the layer when none
existed.

Restore the pre-4.0.1 behaviour of having the per-layer summary report
only the number of failures in that layer.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@
 - Add optional buffering of standard output and standard error during tests,
   requested via the ``--buffer`` option or enabled by default for subunit.
 
+- Fix incorrect failure counts in per-layer summary output, broken in 4.0.1.
+
 
 5.0 (2019-03-19)
 ================
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -400,14 +400,16 @@ def run_tests(options, tests, name, failures, errors, skipped, import_errors):
         t = time.time() - t
         output.stop_tests()
         failures.extend(result.failures)
+        n_failures = len(result.failures)
         if hasattr(result, 'unexpectedSuccesses'):
             # Python versions prior to 2.7 do not have the concept of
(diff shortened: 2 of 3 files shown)


mgedmin:
Oh, so that was a bug?  I noticed it and thought "what a strange design decision, but okay". ;)

Can you please rebase, the changelog is conflicting with git master.

mgedmin:
I'm thinking it's almost time for a new release.  @cjwatson, do you have any other PRs in mind you'd like to get included?

(Perhaps you'd like PyPI access so you could make releases too?)

cjwatson:
I agree.  I'm in the process of doing a full Launchpad test run (which takes several hours) with these changes, by way of a rather more exhaustive test; I'll probably then do another one in subunit mode to make sure that I didn't miss any oddities in the buffering changes.  Maybe early next week then?

We do still have a couple of remaining differences, but nothing I expect to want to submit as PRs at least for now, and the delta is much easier to manage now.

I'd be happy to have PyPI access if that's OK by you.  My username there is 'cjwatson', same as here.

mgedmin:
You have PyPI access now.

We usually make releases with [zest.releaser](http://pypi.org/p/zest.releaser).  No configuration is necessary (other than the usual `~/.pypirc` for twine).  Running `fullrelease` does the changelog/setup.py version updates, git commits, tags, sdist/wheel builds, uploads, post-release version bumps, all with interactive confirmation.

cjwatson:
Done now.  Thanks!

PR #107, opened 2020-06-29
Try out Python 3.9

icemac:
In ZTK tests it breaks.


icemac:
I was misreading the traceback in https://travis-ci.org/github/zopefoundation/zopetoolkit/jobs/703067107. It seems to be `zope.exceptions` which fails. See https://github.com/zopefoundation/zope.exceptions/pull/16

PR #114, opened 2021-03-03
Improve standard stream handling when buffering

cjwatson:
One of the last remaining issues in porting Launchpad to Python 3 is
that our CI infrastructure runs tests with subunit output (thereby
enabling test output buffering), and a few tests assume that on Python 3
they can use `sys.stdout.buffer` or `sys.stderr.buffer` to get an object
suitable for writing binary data to the standard streams.  These tests
currently fail because when zope.testrunner is doing test output
buffering it replaces the standard streams with instances of
`io.StringIO`.

Now, it's true that https://docs.python.org/3/library/sys.html notes
that "the standard streams may be replaced with file-like objects like
`io.StringIO` which do not support the `buffer` attribute"; however,
there are a couple of cases where it's difficult to work around this,
for example because the assumption is baked into existing libraries like
subunit itself.  Besides, I don't think that this note in the standard
library documentation *necessarily* means that zope.testrunner has to
do something inconvenient.

An alternative approach is to use a variant of `io.TextIOWrapper`, which
provides the `buffer` attribute for tests that need it, and can easily
be made to continue to provide the `getvalue` method for internal use.
This seems like a good compromise.

### DIFF
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -898,13 +898,30 @@ class TestResult(unittest.TestResult):
             if hasattr(layer, 'testTearDown'):
                 layer.testTearDown()
 
+    def _makeBufferedStdStream(self):
+        """Make a buffered stream to replace a standard stream."""
+        if sys.version_info[0] >= 3:
+            # The returned stream needs to have a 'buffer' attribute, since
+            # some tests may expect that to exist on standard streams, and a
+            # 'getvalue' method for the convenience of _restoreStdStreams.
+            # This requires some care.
+            class BufferedStandardStream(io.TextIOWrapper):
+                def getvalue(self):
+                    return self.buffer.getvalue().decode(
+                        encoding=self.encoding, errors=self.errors)
+
+            return BufferedStandardStream(
+                io.BytesIO(), newline='\n', write_through=True)
+        else:
+            return StringIO()
+
     def _setUpStdStreams(self):
         """Set up buffered standard streams, if requested."""
         if self.options.buffer:
             if self._stdout_buffer is None:
-                self._stdout_buffer = StringIO()
+                self._stdout_buffer = self._makeBufferedStdStream()
             if self._stderr_buffer is None:
-                self._stderr_buffer = StringIO()
+                self._stderr_buffer = self._makeBufferedStdStream()
             sys.stdout = self._stdout_buffer
             sys.stderr = self._stderr_buffer
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,10 +2,11 @@
  zope.testrunner Changelog
 ===========================
 
-5.2.1 (unreleased)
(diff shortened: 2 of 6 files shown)


mgedmin (approved):
LGTM!

mgedmin:
Uh, we don't have Travis CI any more, and we haven't set up GitHub Actions for zope.testrunner yet, which makes things a bit awkward.

We have scripts and templates in https://github.com/zopefoundation/meta/tree/master/config that automate setting up GitHub Actions, but I haven't used them much.

cjwatson:
Mm, I've used them a couple of times so I'll see if I can get them working for zope.testrunner later today.

mgedmin:
I've made an attempt (#115, #116), but it's trickier than I thought, and I don't mind if somebody else finishes it without waiting for me.

cjwatson:
I'm starting with the bulk job of cleaning up lint errors at the moment.

cjwatson:
#117 is a bit gigantic, but it makes zope.testrunner `flake8`-clean which should make the rest of the conversion easier.

icemac:
Let's to merge this PR wait until #116 is merged.

mgedmin:
#116 was merged.  Would you mind rebasing this on git master?

cjwatson:
Rebased now.  How's this?

src/zope/testrunner/runner.py
```diff
@@ -898,13 +898,30 @@ def testTearDown(self):
...
+            # This requires some care.
+            class BufferedStandardStream(io.TextIOWrapper):
+                def getvalue(self):
+                    return self.buffer.getvalue().decode(
+                        encoding=self.encoding, errors=self.errors)
+
+            return BufferedStandardStream(
+                io.BytesIO(), newline='\n', write_through=True)
+        else:
+            return StringIO()
```
jamadden, line 916: `StringIO` is `from six import StringIO`. On Python 2, that's `StringIO.StringIO`, which is the non-native version (`cStringIO.StringIO` is the native version). 

Long story short, I was concerned about possible performance implications, but then I saw the next lines where we were unconditionally assigning `StringIO` instances, so this shouldn't be a regression. Indeed, on Python 3 it might be an improvement!

jamadden (approved):
LGTM.

src/zope/testrunner/runner.py
cjwatson: Yes, it should be near enough identical on Python 2, and on Python 3 I think `io.StringIO` and an `io.TextIOWrapper` around an `io.BytesIO` are likely to perform similarly enough for our purposes.

I've taken to explicitly using `six.StringIO` or `io.StringIO` in new code rather than doing something like `from six import StringIO`, because the fact that there are multiple kinds of `StringIO` with different semantics is far too confusing (in hindsight it's a shame that the modern version wasn't called `io.TextIO`).  Cleaning that up can be for later though.

PR #116, opened 2021-03-03
Configuring for pure-python

mgedmin:
This does:
- bin/python config-package.py ~/src/zopefoundation/zope.testrunner --type pure-python --with-appveyor --with-pypy --with-docs
- some edits to `.meta.toml` for things that cannot be specified on the command line, like fail_under = 85 and additional flake8/tox/appveyor options
- rerunning bin/python config-package.py again
- adding 3.9 classifier to setup.py to make check-python-versions-happy
- fixes subprocess support on Windows, which I suspect was always broken

It drops some coverage-related configuration, but passes the 85% threshold despite of that:
- we're missing some exclusions from the old .coveragerc
- we're now enabling branch coverage
- we're no longer doing COVERAGE_PROCESS_START (and `coverage.process_startup()` invocations from the setup.py-test-generated wrapper script)

TL;DR: we now use GitHub Actions instead of Travis CI, and we no longer use `setup.py test`, and running layers in a subprocess now works under Windows.


.coveragerc
```diff
@@ -1,20 +0,0 @@
-[run]
-source = zope.testrunner
-parallel = true
-omit =
-    # These files are definitely imported and executed,
-    # but for some reason don't show up as such.
-    # See testrunner-knit
-    src/zope/testrunner/tests/testrunner-ex-pp-products/more/sampletests.py
-    src/zope/testrunner/tests/testrunner-ex-pp-products/sampletests.py
```
mgedmin, line 9: I suppose these need to be moved into `meta.toml` and regenerated into wherever that template puts coverage configs

.coveragerc
```diff
@@ -1,20 +0,0 @@
...
-[report]
-precision = 2
-exclude_lines =
-    pragma: no cover
-    if __name__ == '__main__':
-    raise NotImplementedError
-    self.fail
-    raise AssertionError
-    if is_jython:
-    if sys.platform == 'win32':
```
mgedmin, line 20: This also I'm sure includes things not mentioned in the standard template.

.meta.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[python]
+with-appveyor = true
+with-pypy = true
+with-legacy-python = true
+with-docs = false
+with-sphinx-doctests = false
+
+[coverage]
+fail-under = 0
```
mgedmin, line 15: This needs fixing

.travis.yml
```diff
@@ -1,25 +0,0 @@
...
-env:
-  - SUBUNIT=true
-  - SUBUNIT=false
-install:
-    - pip install -U pip setuptools
-    - pip install -U coverage coveralls
-    - pip install -U -e .[test]
-    - test "$SUBUNIT" != true || pip install -U -e .[subunit]
-script:
-    - COVERAGE_PROCESS_START=`pwd`/.coveragerc coverage run setup.py -q test -q
```
mgedmin, line 19: Oof, I suppose this will need moving into tox.ini

appveyor.yml
```diff
@@ -1,7 +1,7 @@
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/pure-python
 environment:
 
-  SUBUNIT: 0
-
```
mgedmin, line 4: This will need to be moved into the template

appveyor.yml
```diff
@@ -13,29 +13,30 @@ environment:
...
-    - { python: 27, SUBUNIT: 1 }
-    - { python: 27-x64, SUBUNIT: 1 }
-    - { python: 35, SUBUNIT: 1 }
-    - { python: 35-x64, SUBUNIT: 1 }
-    - { python: 36, SUBUNIT: 1 }
-    - { python: 36-x64, SUBUNIT: 1 }
-    - { python: 37, SUBUNIT: 1 }
-    - { python: 37-x64, SUBUNIT: 1 }
-    - { python: 38, SUBUNIT: 1 }
-    - { python: 38-x64, SUBUNIT: 1 }
```
mgedmin, line 25: These will need to be resurrected somehow

appveyor.yml
```diff
@@ -13,29 +13,30 @@ environment:
...
+      if (-not (Test-Path $env:PYTHON)) {
+        curl -o install_python.ps1 https://raw.githubusercontent.com/matthew-brett/multibuild/11a389d78892cf90addac8f69433d5e22bfa422a/install_python.ps1
+        .\install_python.ps1
+      }
+  - ps: if (-not (Test-Path $env:PYTHON)) { throw "No $env:PYTHON" }
   - echo "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /x64 > "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\bin\amd64\vcvars64.bat"
-  - python -m pip install -U setuptools pip
-  # NB: pip install -e .[test] fails for obscure namespace package reasons
-  - pip install -U .[test]
-  - if %SUBUNIT%==1 pip install -U .[subunit]
```
mgedmin, line 33: Oof this too

appveyor.yml
```diff
@@ -13,29 +13,30 @@ environment:
...
+  - pip install -U setuptools wheel
+  - pip install -U -e .[test]
 
 build: false
+matrix:
+  fast_finish: true
 
 test_script:
-  - python setup.py -q test -q
+  - zope-testrunner --test-path=src
```
mgedmin, line 39: Hmm.  Can we rely on zope.testrunner to run its own tests without relying on whatever hacks `setup.py test` does?

tox.ini
```diff
@@ -1,35 +1,69 @@
...
+    lint
+    py27
+    py35
+    py36
+    py37
+    py38
+    py39
+    pypy
+    pypy3
+    coverage
```
mgedmin, line 15: Yeah this drops docs and all the subunit variants, not good.

tox.ini
```diff
@@ -1,35 +1,69 @@
...
 
 [testenv]
+usedevelop = true
+deps =
+    zope.testrunner
+commands =
+    zope-testrunner --test-path=src {posargs:-vc}
 extras =
     test
-    subunit: subunit
```
mgedmin, line 8: We need to keep this bit.

tox.ini
```diff
@@ -1,35 +1,69 @@
...
+    check-python-versions
 
 [testenv:coverage]
-usedevelop = true
-basepython =
-    python3
-commands =
-    coverage run setup.py -q test -q
-    coverage combine
-    coverage report --fail-under=85
```
mgedmin, line 19: Number!

tox.ini
```diff
@@ -1,35 +1,69 @@
...
-setenv =
-    COVERAGE_PROCESS_START = {toxinidir}/.coveragerc
```
mgedmin, line 21: We probably need to keep this

tox.ini
```diff
@@ -1,35 +1,69 @@
...
-extras =
-    test
-    subunit
```
mgedmin, line 24: and this

tox.ini
```diff
@@ -1,35 +1,69 @@
...
 deps =
     coverage
-
-[testenv:docs]
-basepython =
-    python3.6
+    coverage-python-version
+    zope.testrunner
 commands =
-    sphinx-build -b html -d docs/_build/doctrees docs docs/_build/html
```
mgedmin, line 32: And I totally forgot to enable sphinx docs when running the meta script.

mgedmin (commented):
So this is a larger job than I initially realized.

I'm mostly worried about the chicken-and-egg situation of testing zope.testrunner using zope.testrunner.  Can we drop setup.py test or does it do something essential to enable bootstrapping?

jamadden:
Is zope.testrunner special enough that it's not worth it to "standardize" with this tool? Setting up a GHA .yml file that replicates what .travis.yml did for pure-python projects is relatively straightforward-ish  (not that I'm volunteering right now 😄), though of course the more complicated .travis.yml is, the worse the GHA will always be.

cjwatson:
It's not really a chicken-and-egg question given that `setup.py test` uses `zope.testrunner` internally anyway, just a question of whether its extra hacks matter.

I think we should give removing it a try.  I killed off just `setup.py test` locally without doing any of the rest of the stuff in this PR (except for updating `tox.ini`) and `tox` seems happy locally except for maybe a problem with insufficient coverage; if there are still any CI problems with that then we should probably investigate them harder.

icemac:
When #117 is merged GHA should at least run the tests.

.meta.toml
```diff
@@ -0,0 +1,15 @@
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/pure-python
+[meta]
+template = "pure-python"
```
icemac, line 4: Maybe the template `buildout-recipe` is better suited here as it provides infrastructure for coverage when using multiple processes.

mgedmin: Nah, zope.testrunner is not a buildout recipe.  I'd rather add a flag to the pure-python recipe for supporting multiprocess coverage, I expect it'll come in handy for other packages (e.g. zdaemon).

cjwatson:
> When #117 is merged GHA should at least run the tests.

This is done now, so worth rebasing I think.

mgedmin:
PR updated, still very not ready.

appveyor.yml
mgedmin: The consensus seem to be "probably"?

tox.ini
mgedmin: Semi-fixed, but I'm now enumerating all the supported Python versions in .meta.toml, in `additional-envlist`.

tox.ini
mgedmin: Fix relies on https://github.com/zopefoundation/meta/pull/71.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,58 @@
...
+        - ["2.7",   "py27"]
+        - ["3.5",   "py35"]
+        - ["3.6",   "py36"]
+        - ["3.7",   "py37"]
+        - ["3.8",   "py38"]
+        - ["3.9",   "py39"]
+        - ["pypy2", "pypy"]
+        - ["pypy3", "pypy3"]
+        - ["3.8",   "docs"]
+        - ["3.8",   "coverage"]
```
mgedmin, line 28: Oops, need to add all the subunit variants here.

setup.cfg
```diff
@@ -1,7 +1,13 @@
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/pure-python
 [bdist_wheel]
 universal = 1
 
 [flake8]
+doctests = 1
```
mgedmin, line 7: Oh, I've only just now realized the additional flake8 errors I had to additionally fix came from `doctests = 1` here!

tox.ini
```diff
@@ -1,43 +1,81 @@
...
-    subunit: subunit
-commands =
-    python setup.py -q test -q
-
-[testenv:coverage]
 usedevelop = true
-basepython =
-    python3
+deps =
+    zope.testrunner
```
mgedmin, line 22: I wonder if we should drop this?  Kind of pointless to install zope.testrunner from pypi before reinstalling it from the local source tree.

(This would need another template tweak.)

mgedmin (commented):
This no longer feels a hopeless task!

tox.ini
mgedmin: this could also explain the coverage percentage drop!

mgedmin: huh, re-enabling the `subunit` extra for tox -e coverage brought our percentage back to 85%!

mgedmin:
I would prefer to get https://github.com/zopefoundation/meta/pull/71 merged before I create a new PR for tweaking appveyor.yml.

There currently is an escape hatch, but I don't want to wrap the entire appveyor.yml inside a TOML list, I only need a few tweaks (global envvars, and additional matrix rows)

mgedmin:
Ok, current status: appveyor is broken as heck, and I'm not sure why.

.meta.toml
icemac: At least in the case of buildout recipes the coverage config has to be stored in a `.coveragerc` – that's the main difference to the `pure-python` template. Having both `.coveragerc` and coverage config in `tox.ini` in a template makes it even more complex.
Maybe the name of the `buildout-recipe` template is only too specific and it should be called `coverage-over-subprocesses` (?).

tox.ini
icemac: I think it would be worth the additional template tweak.

mgedmin:
Ok, so the Windows breakage: zope-testrunner cannot run tests in a subprocess because it tries to execute `/python27/python.exe /python27/scripts/zope-testrunner`, and python.exe says `/python27/scripts/zope-testrunner` does not exist.

Before, we had this interesting bit:
https://github.com/zopefoundation/zope.testrunner/blob/9b7cc179ee857e434f85d7e1cdc873ef4ded1bd0/src/zope/testrunner/__init__.py#L63-L64

i.e. I think we've been running `/python27/python.exe -c "from zope.testrunner import run; run()"`.

I wonder if we could always do [sys.executable, `-mzope.testrunner`] instead?

mgedmin:
I don't have a Windows machine to test, or, rather, I\'m not going to dual-boot into Windows on this laptop.  IIRC the 'scripts/' folder gets three wrapper scripts per console_script entry, one with .py, one with .exe, and the third one is some weird  -script.exe?  Or -wrapper.exe?  I honestly don't remember.

I like the idea of doing `python -m zope.testrunner` instead of trying to figure out what's the right wrapper script filename to pass to python.exe.  I'm only going to do that if the current logic looking for the wrapper script fails to find one, so this should not regress on Linux or MacOS.  And AFAIU things on Windows were already broken wrt subprocesses, except for that one case where you were using `python setup.py test` as a wrapper for zope.testrunner.

So, I think this is finally in a mergeable state.  I don't have the energy to see if I can improve coverage numbers by tweaking coverage configs.  I might find the energy to tweak the meta config template to drop the zope.testrunner self-dependency in tox.ini, especially if somebody gently prods me to do it.

mgedmin:
It seems that I ran out of steam for the nice-to-have improvements (deps in tox.ini, investigating coverage numbers).

I propose merging this as-is, unless anyone objects?.

icemac:
This PR creates a much better situation than the one we currently have on `master`. Maybe you could create an issue so we do not forget the nice-to-have improvements and can tackle them later on.

tox.ini
mgedmin: https://github.com/zopefoundation/meta/pull/85

mgedmin:
> Maybe you could create an issue so we do not forget the nice-to-have improvements and can tackle them later on.

https://github.com/zopefoundation/zope.testrunner/pull/118

PR #117, opened 2021-03-03
Clean up lots of lint and enable flake8

cjwatson:
I added a few things to `per-file-ignores` where it seemed that making
`flake8` happy with them directly would involve making the code less
readable or idiomatic, but most of this is just an enormous pile of
whitespace changes and similar.

In practice there should be no functional change as a result of this
commit.  The only things I can think of seem extremely unlikely, and can
only affect Python 2:

 * On Python 2, changing bare `except:` to `except BaseException:` could
   conceivably break somebody relying on being able to raise an instance
   of an old-style class as an exception (but this will break on Python
   3 anyway);

 * Changing `if t is types.InstanceType:` to `if issubclass(t,
   types.InstanceType):` could perhaps change behaviour for some very
   strange code (but `TrackRefs.update` can't possibly work on Python 3
   at the moment anyway, since `types.InstanceType` doesn't exist
   there).

All `tox` environments pass locally.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,8 @@
 [bdist_wheel]
 universal = 1
+
+[flake8]
+per-file-ignores =
+    src/zope/testrunner/tests/test_runner.py: E701
+    src/zope/testrunner/tests/testrunner-ex/sample2/sampletests_ntds.py: E702
+    src/zope/testrunner/tests/testrunner-ex/sample3/sampletests_d.py: E702
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,6 +3,10 @@ universal = 1
 
 [flake8]
 per-file-ignores =
+    # E701: multiple statements on one line (colon)
+    # TestLayerOrdering uses this for compact declaration of class graphs.
     src/zope/testrunner/tests/test_runner.py: E701
+    # E702: multiple statements on one line (semicolon)
+    # Some tests use the "import pdb; pdb.set_trace()" idiom.
     src/zope/testrunner/tests/testrunner-ex/sample2/sampletests_ntds.py: E702
     src/zope/testrunner/tests/testrunner-ex/sample3/sampletests_d.py: E702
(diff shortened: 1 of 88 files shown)


setup.cfg
```diff
@@ -1,2 +1,8 @@
 [bdist_wheel]
 universal = 1
+
+[flake8]
+per-file-ignores =
```
icemac, line 5:
```suggestion
# E701 Multiple statements on one line (colon)
# E702 Multiple statements on one line (semicolon) 
per-file-ignores =
```

It would be nice to document what the codes mean.

src/zope/testrunner/tests/test_doctest.py
```diff
@@ -151,13 +159,15 @@
...
+                    r'[^\n]+\n[^\n]+\n',
                     re.MULTILINE),
          r''),
-        (re.compile(r'^{\w+} +File "{\w+}[^\n]*(doctest|unittest|case)(/__init__)?.py{\w+}", [^\n]+\n[^\n]+\n',
+        (re.compile(r'^{\w+} +File "{\w+}[^\n]*(doctest|unittest|case)'
+                    r'(/__init__)?.py{\w+}", [^\n]+\n[^\n]+\n',
                     re.MULTILINE),
          r''),
-        (re.compile('import pdb; pdb'), 'Pdb()'), # Py 2.3
+        (re.compile('import pdb; pdb'), 'Pdb()'),  # Py 2.3
```
icemac, line 170: Really Python 2.3?

icemac (approved):
LGTM besides one suggestion.

Tests ran fine locally. I'd suggest to merge this ticket before #116, so at least the flake8 tests run successfully over there.

mgedmin:
I'm going to LGTM this without looking at the diff.  Maybe FGTM (Feels Good To Me) would be more appropriate?

(I did run tox -p auto in this branch, and saw that all was good.)

setup.cfg
cjwatson: Done, but I added some more detailed reasoning as well.

src/zope/testrunner/tests/test_doctest.py
cjwatson: Well, quite - I noticed that, but didn't think it would be right to change it in this already packed PR.

PR #118, opened 2021-03-16
Configuring for pure-python

mgedmin:
Newer template: builds all branches, drops self-dependency on zope.testrunner in tox.ini.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,6 @@ name: tests
 
 on:
   push:
-    branches: [ master ]
   pull_request:
   schedule:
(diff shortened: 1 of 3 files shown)


jamadden (approved):
LGTM

cjwatson (approved):
LGTM.  While you're here, I wonder if you could add a note to CHANGES.rst about adding support for Python 3.9?  Your previous CI work had that effect and it's something we'd ordinarily note in the changelog - I only noticed that it wasn't there after it landed.

mgedmin:
It's probably time to release 5.3.0?

icemac:
> It's probably time to release 5.3.0?

I think this is a good idea.

PR #122, opened 2021-04-28
do not disable existing loggers during logsupport initialization

### LINKED ISSUE #120: Logsupport can interfere with tests involving logging
I am currently analyzing `ZEO` problems under Python 2. To this end, I have cloned the `ZEO` repository and have it integrated into my virtual env with `python setup.py develop`. When I run `zope-testrunner --test-path src -t commit_from_wrong_partition` these tests fail.
The tests use logging to verify the required actions:
```pycon
    >>> import logging
    >>> logger = logging.getLogger('ZODB.blob.copied')
    >>> handler = logging.StreamHandler(sys.stdout)
    >>> logger.propagate = False
    >>> logger.setLevel(logging.DEBUG)
    >>> logger.addHandler(handler)
...
    >>> transaction.commit() # doctest: +ELLIPSIS
    Copied blob file ...
```
The tests fail because the used logger has been disabled during `zope.testrunner`'s `logsupport.Logging` initialization: `ZEO` has a `log.ini` and `logging.config.fileConfig` disables all existing loggers not configured via the configuration file.

The tests do not fail under Python 3 even though the same logic is used. Apparently, they do not always fail under Python 2 (e.g. the tests succeed when run on `githup`) -- obviously, whether or not the logger used by the tests exists during `Logging` initialization depends on fine details.

`logging.config.fileConfig` has the parameter `disable_existing_loggers` (default `True`). We should consider whether we call it with `disable_existing_loggers=False` to make tests involving logging more reliable.

d-maurer:
fixes #120.

`logsupport.Logging` used to call `logging.config.fileConfig` without the `disable_existing_loggers` parameter (which defaults to `True`). This caused to disable loggers implicitly created during earlier initialization steps. When a test used one of those loggers, the test could fail (because it does not expect the logger to be disabled).

This PR calls `fileConfig` with `disable_existing_loggers=False`. All loggers existing at that time have been implicitly created in earlier `testrunner` initialization steps and therefore come from common libraries; it properly designed, they will not have associated handlers and defer to higher up loggers under application configuration control. Therefore, it does not harm to leave them enabled.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@
 5.3.1 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Do not disable existing loggers during logsupport initialization
+  (`#120 <https://github.com/zopefoundation/zope.testrunner/pull/120>`_).
+
 
 
 5.3.0 (2021-03-17)
--- a/src/zope/testrunner/logsupport.py
+++ b/src/zope/testrunner/logsupport.py
@@ -39,7 +39,7 @@ class Logging(zope.testrunner.feature.Feature):
 
         logini = os.path.abspath("log.ini")
         if os.path.exists(logini):
-            logging.config.fileConfig(logini)
+            logging.config.fileConfig(logini, disable_existing_loggers=False)
         else:
             # If there's no log.ini, cause the logging package to be
             # silent during testing.


mgedmin (approved):
Words cannot express how much I approve this!

d-maurer:
Marius Gedminas wrote at 2021-4-28 00:11 -0700:
***@***.*** approved this pull request.
>
>Words cannot express how much I approve this!

Thank you for your review and the kind words!

PR #123, opened 2021-04-29
improve `--help` documentation for `--package-path` option

### LINKED ISSUE #121: `package-path` option supoptimally descibed by `--help`
`zope-testrunner --help` reports for option `package-path`: `--package-path PACKAGE_PATH PACKAGE_PATH`. `--package-path PACKAGE_PATH PACKAGE_NAME` would be more appropriate (but is potentially not directly supported by `argparse`).

d-maurer:
fixes #121.

This PR (slightly) improves the documentation for option `--package-path` output by `-h/--help` by specifying `metavar="ARG"` and some rewording of the usage text.

### DIFF
--- a/src/zope/testrunner/options.py
+++ b/src/zope/testrunner/options.py
@@ -418,14 +418,14 @@ will be run.
 
 setup.add_argument(
     '--package-path', action="append", dest='package_path', nargs=2,
+    metavar="ARG",
     help="""\
 Specify a path to be searched for tests, but not added to the Python
 search path.  Also specify a package for files found in this path.
 This is used to deal with directories that are stitched into packages
 that are not otherwise searched for tests.
 
-This option takes 2 arguments.  The first is a path name. The second is
-the package name.
+This option takes 2 arguments specifying the path and the package.
 
 This option can be used multiple times to specify
 multiple search paths.  The path is usually specified by the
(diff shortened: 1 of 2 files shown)


src/zope/testrunner/options.py
```diff
@@ -418,14 +418,14 @@ def _regex_search(s):
...
+    metavar="ARG",
     help="""\
 Specify a path to be searched for tests, but not added to the Python
 search path.  Also specify a package for files found in this path.
 This is used to deal with directories that are stitched into packages
 that are not otherwise searched for tests.
 
-This option takes 2 arguments.  The first is a path name. The second is
-the package name.
+This option takes 2 arguments specifying the path and the package.
```
mgedmin, line 428: BTW do you have any idea what this option actually does?  Because I don't, and the text isn't helping.  Maybe an example would help?

mgedmin (approved):
LGTM

d-maurer:
Marius Gedminas wrote at 2021-4-29 02:45 -0700:
> ...
>BTW do you have any idea what this option actually does?  Because I don't, and the text isn't helping.  Maybe an example would help?

The first argument to `--package-path` has the same effect
as `--test-path`, i.e. tests are searched below the specified path.
The second argument specifies the package associated with this
path. This information may be necessary in order to resolve
relative imports correctly.

Example from `ZEO` tests (in a `ZEO` checkout directory):
`zope-testrunner --package-path src/ZEO/asyncio ZEO.asyncio`
runs the tests in `src/ZEO/asyncio`.


Thanks for your review.

PR #124, opened 2021-05-10
Adapt to python 3.10 SyntaxError messages

jamesjer:
SyntaxError messages are becoming more verbose in python 3.10.  See https://docs.python.org/3.10/whatsnew/3.10.html#better-error-messages.  As a result, 4 tests fail with the current python 3.10 beta.  See https://bugzilla.redhat.com/show_bug.cgi?id=1958894 for the Fedora bug report.

With this change, the tests pass with both python 3.9 and 3.10 beta 1.

### DIFF
--- a/src/zope/testrunner/tests/testrunner-debugging-import-failure.rst
+++ b/src/zope/testrunner/tests/testrunner-debugging-import-failure.rst
@@ -41,7 +41,7 @@ Post-mortem debugging also works when there is an import failure.
       File ".../TESTS-DIR/tests.py", line 2
         impot doctest
                     ^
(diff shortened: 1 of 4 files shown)


mgedmin:
The py27 error looks unrelated to the changes made in this PR.  I fear there's some nondeterminism from running test layers in parallel.

I'm not opposed to merging forward-looking Python 3.10 fixed before we have a Python 3.10 row in the CI matrix.  This is the sort of fix that's unlikely to regress.

icemac:
Thank you for this PR and for looking into it. 😀

PR #126, opened 2021-07-27
Fix tests with testtools >= 2.5.0

### LINKED ISSUE #125: subunit tests started failing recently
With no changes to zope.testrunner's codebase itself, weekly GitHub Actions builds started failing recently.  Example failure:  https://github.com/zopefoundation/zope.testrunner/runs/3097229321?check_suite_focus=true

The failures all involve subunit.

cjwatson:
testtools 2.5.0 only supports Python 3, and so switched from traceback2
to the standard library's traceback.  However, this includes the fix for
https://bugs.python.org/issue24695, and so doesn't produce a traceback
header when there's no traceback, in particular for `SyntaxError`
exceptions.  Adjust the tests to tolerate this.

Fixes #125.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 - Do not disable existing loggers during logsupport initialization
   (`#120 <https://github.com/zopefoundation/zope.testrunner/pull/120>`_).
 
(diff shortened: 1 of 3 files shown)


mgedmin (approved):
Thank you!

PR #128, opened 2022-03-04
Fix GHA PyPy3 test runs on Windows

icemac:
See https://github.com/zopefoundation/zope.testrunner/runs/5350155648?check_suite_focus=true as an example for a broken test run.


cjwatson (approved):

icemac:

PR #129, opened 2022-03-09
Inline a small part of random.Random.shuffle

### LINKED ISSUE #119: random.shuffle(random=...) is deprecated in 3.9
Python 3.9 test runs (e.g. https://github.com/zopefoundation/zope.testrunner/runs/2129207371) show a number of instances of this:

```
/home/runner/work/zope.testrunner/zope.testrunner/src/zope/testrunner/shuffle.py:56: DeprecationWarning: The *random* parameter to shuffle() has been deprecated
since Python 3.9 and will be removed in a subsequent version.
  rng.shuffle(tests, random=rng.random)
```

See commit b7e1944cb3 for some of the history here.  Maybe we just need to inline the relevant code from `random.shuffle` or something?

cjwatson:
This was deprecated in Python 3.9 and removed in 3.11.

Fixes #119.


mgedmin (approved):
lgtm

I would mildly prefer extracting the loop into its own top-level helper function (`random_shuffle`), and maybe converting the long comment into a docstring.

(The `floor = math.floor` bit also feels a bit of excessive, this is not a hot spot for zope.testrunner.)

icemac:
@cjwatson What is the current status of this PR?

PR #134, opened 2022-05-19
logging configuration via envvar `ZOPE_TESTRUNNER_LOG_INI`

d-maurer:
`zope.testrunner` allows to specify a logging configuation in the file `log.ini` of the current working directory. This works fine until spawned subprocesses come into play: the spawn creates a pristine process image and the logging configuration is lost (see e.g. https://github.com/zopefoundation/ZEO/pull/195#issuecomment-1130260432 for the problems this can make).

The ZEO tests plagued by missing child process log messages often change the current working directory. Then, they have difficulties to locate the original logging configuration used by `zope.testrunner` (because the original current working directory is not easily accessible).
This PR therefore allows to specify the file path for the logging configuration via the envvar `ZOPE_TESTRUNNER_LOG_INI` (child processes can easily access it). `zope.testrunner` stores the absolute path of the logging configuration file it has used in this envvar -- for potential use by child processes.


dataflake (commented):
This works, I'm just concerned that if it's not an easily discoverable command line switch then it may have very little visibility, people may not know it's there.

d-maurer:
Jens Vagelpohl wrote at 2022-5-19 08:05 -0700:
***@***.*** commented on this pull request.
>
>This works, I'm just concerned that if it's not an easily discoverable command line switch then it may have very little visibility, people may not know it's there.

I agree that an explicit logging parameter would be better
than the "log.ini inside working directory" or "envvar".
My `zope.testrunner` wrapper (`dm.zopepatches.ztest`) has such
a parameter.
To make things worse, the logging support is likely nowhere documented.

The PR does nothing regarding this state.
The introduced envvar is mainly there to get an easy way
to communicate logging information to child processes
(such that they can recreate the logging configuration if they care about it).
For users, it is sufficient to know about the `log.ini`;
the envvar is set automatically.

Loggging support is implemented as a `Feature`.
I do not know whether features have access to the options.
I fear that they are prior to options (to allow them
to add new parameters).
Letting an option control the logging would require some
research around the "feature" concept.

icemac (approved):
LGTM – It improves the situation, so I like it.

dataflake (approved):

PR #135, opened 2022-06-03
use `sys._current_frames` to detect new threads

### LINKED ISSUE #130: "New threads" information unreliable for threads started with `thread.start_new_thread`
`zope.testrunner` tries to check whether a test has left new threads behind. For this, it uses `threading.enumarate`. However, `threading` may or may not know about threads started with `thread.start_new_thread`. If it does know, it does not know whether the thread has stopped.

As a consequence, `zope.testrunner` may miss threads (if `threading` does not know them) or wrongly report a left behind thread (if `threading` knows it but has not been informed that is has stopped).

I suggest to implement the "New threads" feature via `sys._current_frame` (rather than `threading.enumerate`).

d-maurer:
fixes #130.

This PR uses `sys._current_frames` (if available) as base for new thread detection.

In contrast to the formerly used `threading.enumerate`, `_current_frames` has reliable information about threads started via `_thread.start_new_thread`. This makes the runner's "new thread" information reliable even if  threads are started this way.

### DIFF
--- a/src/zope/testrunner/threadsupport.py
+++ b/src/zope/testrunner/threadsupport.py
@@ -33,7 +33,7 @@ else:
         th_known = dict((t.ident, t) for t in threading.enumerate())
         return [ThreadProxy(th_known[i] if i in th_known else DummyThread(i))
                 for i in running]
-    
+
 
 class ThreadProxy(object):
     """auxiliary class to provide ident based ``__eq__``."""
@@ -61,5 +61,3 @@ class DummyThread(object):
 
     def is_alive(self):
         return True
-
-
--- a/src/zope/testrunner/threadsupport.py
+++ b/src/zope/testrunner/threadsupport.py
@@ -1,0 +1,65 @@
+##############################################################################
+#
+# Copyright (c) 2022 Zope Foundation and Contributors.
+# All Rights Reserved.
+#
+# This software is subject to the provisions of the Zope Public License,
+# Version 2.1 (ZPL).  A copy of the ZPL should accompany this distribution.
+# THIS SOFTWARE IS PROVIDED "AS IS" AND ANY AND ALL EXPRESS OR IMPLIED
+# WARRANTIES ARE DISCLAIMED, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF TITLE, MERCHANTABILITY, AGAINST INFRINGEMENT, AND FITNESS
+# FOR A PARTICULAR PURPOSE.
+#
+##############################################################################
+"""Thread support beyond ``threading``.
+
+``threading.enumerate`` may or may not know threads
+started with ``_thread.start_new_thread``. If it knows them,
+it does not know when they stop.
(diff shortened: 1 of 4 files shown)


icemac (approved):
Looks good from reading through the code. But I have to admit that I do not understand it in detail.

d-maurer:
Michael Howitz wrote at 2022-6-10 08:35 -0700:
***@***.*** approved this pull request.
>
>Looks good from reading through the code. But I have to admit that I do not understand it in detail.

It replaces `threading.enumerate` by (the locally implemented)
`threadsupport.enumerate`.

`threading.enumerate` knows about all `Thread` instances and their fate.
It knows about threats created via `start_new_thread` only if
they happen to call `current_thread` (e.g. because they issue a logging
call); it does not know when those threads have terminated -- it thinks
they run forever. Therefore, the "new thread" information of
`zope.testrunner` is unreliable for threads created with
`start_new_threads`: it may not detect them at all (if they
did not happen to call `current_thread`) or may report them as
not finished.

`threadsupport.enumerate` uses `sys._current_frames` -- a dict
mapping thread ids to the current frame stack of the thread.
It knows reliably all currently existing threads.
`threadsupport.enumerate` does not return the thread objects directly
but wraps them in order to ensure that objects with the same
thread id are considered equal.
The wrapper delegates method calls to the underlying thread object.
`threadsupport.enumerate` uses the thread description of
`threading` if it exists for a thread; otherwise, it creates a dummy thread
object.

icemac:
@d-maurer Do you think we should cut a new release with the changes you made lately?

d-maurer:
Michael Howitz wrote at 2022-6-23 23:01 -0700:
***@***.*** Do you think we should cut a new release with the changes you made lately?

I have no objections.

icemac:
I just released https://pypi.org/project/zope.testrunner/5.5/.

d-maurer:
Michael Howitz wrote at 2022-6-24 08:28 -0700:
>I just released https://pypi.org/project/zope.testrunner/5.5/.

Thank you!

src/zope/testrunner/threadsupport.py
```diff
@@ -0,0 +1,63 @@
...
+                for i in running]
+
+
+class ThreadProxy(object):
+    """auxiliary class to provide ident based ``__eq__``."""
+    def __init__(self, thread):
+        self.thread = thread
+
+    def __eq__(self, other):
+        return self.thread.ident == other.thread.ident
```
mgedmin, line 44: Are we still supporting Python 2.7 in zope.testrunner?   On 2.7 you need to manually define a custom `__ne__` if you define a custom `__eq__`.

d-maurer:
Marius Gedminas wrote at 2022-9-10 08:23 -0700:
***@***.*** commented on this pull request.
>Are we still supporting Python 2.7 in zope.testrunner?   On 2.7 you need to manually define a custom `__ne__` if you define a custom `__eq__`.

The class has very limited use (well checked by the tests).
The missing `__ne__` is therefore likely not a problem.

PR #140, opened 2022-12-06
Fix GHA: ubuntu-latest no longer contains Python 2.7 up to 3.6

icemac:
* Add support for Python 3.11.
* isort imports


icemac:
Sorry for such a big PR, it got more and more which was needed to get tests green. The only thing I actually wanted was to update with current `meta/config` to get legacy Python tests green again.

d-maurer (approved):

.meta.toml
```diff
@@ -58,10 +60,11 @@ additional-config = [
...
-    "- [\"3.10\",   \"py310-subunit\"]",
-    "- [\"pypy2\", \"pypy-subunit\"]",
+    "- [\"3.10\",  \"py310-subunit\"]",
+    "- [\"3.11\",  \"py311-subunit\"]",
+    "- [\"pypy-2.7\", \"pypy-subunit\"]",
     "- [\"pypy-3.7\", \"pypy3-subunit\"]",
     ]
 additional-exclude = [
-    "- { os: windows, config: [\"pypy2\", \"pypy-subunit\"] }",
+    "- { os: [\"windows\", \"windows-latest\"], config: [\"pypy-2.7\", \"pypy-subunit\"] }",
```
mgedmin, line 69: (This is why I hate TOML as a format.)

src/zope/testrunner/formatter.py
```diff
@@ -788,12 +790,10 @@ def id(self):
...
-    from testtools.content import (
-        Content,
-        ContentType,
-        content_from_file,
-        text_content,
-    )
+    from testtools.content import Content
+    from testtools.content import ContentType
+    from testtools.content import content_from_file
+    from testtools.content import text_content
```
mgedmin, line 796: Oh, does our isort configuration mandate one import per line?

Why are we doing that to ourselves?

src/zope/testrunner/tests/testrunner-arguments.rst
```diff
@@ -40,28 +40,28 @@ different default, that one takes effect.
...
-      test_3 (sample1.sampletests.test111.TestNotMuch)
+      test_x1 (sample1.sampletests.test111.TestA...)
+      test_y0 (sample1.sampletests.test111.TestA...)
+      test_z0 (sample1.sampletests.test111.TestA...)
+      test_x0 (sample1.sampletests.test111.TestB...)
+      test_y1 (sample1.sampletests.test111.TestB...)
+      test_z0 (sample1.sampletests.test111.TestB...)
+      test_1 (sample1.sampletests.test111.TestNotMuch...)
+      test_2 (sample1.sampletests.test111.TestNotMuch...)
+      test_3 (sample1.sampletests.test111.TestNotMuch...)
```
mgedmin, line 51: I don't understand this change.

The commit says it's needed to make things compatible with Python 3.11.  What does Python 3.11 produce for the test names?

Oh, it repeats the test name inside the parentheses, like

> test_something (sampletests_buffering.TestSomething1.test_something)

Couldn't we have fixed that with a renormalizer?  Adding a bazillion `...`s to all the test files seems like a pain.  Of course that's a past pain now, and changing everything back to use a renormalizer would be even more pain.

tox.ini
```diff
@@ -30,15 +31,25 @@ extras =
...
     check-python-versions >= 0.19.1
     wheel
+    flake8
+    isort
+
+[testenv:isort-apply]
+basepython = python3
+commands_pre =
+deps =
+    isort
```
mgedmin, line 50: This (and other tool-oriented testenvs) should have a `skip_install = true` because we don't need to pip install zope.testrunner to run isort/flake8/etc.

This is more of a self-comment to go and file the same issue against the meta/config repo.

mgedmin (approved):
This looks like it was a tremendous job, and it makes me very happy to see someone caring about zope.testrunner.

LGTM.

(I'd've preferred an alternative approach to fixing Python 3.11 test names, but it'll do.)

src/zope/testrunner/formatter.py
icemac: I don't know. I am just blindly using `tox -eisort-apply`.
The rules are defined here: https://github.com/zopefoundation/meta/blob/d8ee00613a60c98a8d2227b0564403d0e447741a/config/default/setup.cfg.j2#L49-L61 – I personally use different rules (no groups of imports, just sort all imports alphabetically) but I do not care as long there is a tool to order the imports as defined by the rules.

src/zope/testrunner/tests/testrunner-arguments.rst
icemac: I was thinking about a renormalizer but did not find a good solution: The test names are too different to be normalized easily. And doctests do not yet seem to repeat the test name in the output.

tox.ini
icemac: Fixed in https://github.com/zopefoundation/zope.testrunner/pull/140/commits/b334aff904d33fda4d8e5a110a0288c81d1680ca

PR #142, opened 2022-12-06
Don't trigger post-mortem debugger for skipped tests

### LINKED ISSUE #141: running with `-D` triggers debugger for skipped tests
### What I did:

Run tests for a package with skipped tests, like this `tests/test_m.py`:

```py
import unittest

class Test1(unittest.TestCase):

    @unittest.skip('skipped !')
    def test(self):
        self.fail('1')


class Test2(unittest.TestCase):

    def test(self):
        self.assertEqual(1, 2)

```

with
```
zope-testrunner -D --path . --test-path . 
```

### What I expect to happen:

`Test1` is skipped, `Test2` fails and the post mortem debugging starts.

### What actually happened:

The post mortem debugger started on the exception raised internally by unittest to skip the test: 

```
Running zope.testrunner.layer.UnitTests tests:
  Set up zope.testrunner.layer.UnitTests in 0.000 seconds.


Error in test test_skipped (tests.test_m.Test1)
  File "/opt/slapgrid/d02ebd529f47acc392f6b079c09d3e4b/parts/python3/lib/python3.8/unittest/case.py", line 741, in debug
    getattr(self, self._testMethodName)()
  File "/opt/slapgrid/d02ebd529f47acc392f6b079c09d3e4b/parts/python3/lib/python3.8/unittest/case.py", line 120, in skip_wrapper
    raise SkipTest(reason)
unittest.case.SkipTest: skipped !

unittest.case.SkipTest: skipped !

> /opt/slapgrid/d02ebd529f47acc392f6b079c09d3e4b/parts/python3/lib/python3.8/unittest/case.py(120)skip_wrapper()
-> raise SkipTest(reason)
(Pdb) q
Tearing down left over layers:
  Tear down zope.testrunner.layer.UnitTests in 0.000 seconds.
```
### What version of Python and Zope/Addons I am using:

zope.testrunner master branch

perrinjerome:
Fixes #141

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@
   Python 3.9 and removed in 3.11 (`#119
   <https://github.com/zopefoundation/zope.testrunner/issues/119>`_).
 
+- Don't trigger post mortem debugger for skipped tests. ( `#141
+  <https://github.com/zopefoundation/zope.testrunner/issues/141>`_).
+
 
 5.5.1 (2022-09-07)
 ==================
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -384,6 +384,8 @@ def run_tests(options, tests, name, failures, errors, skipped, import_errors):
                         test.debug()
                     except KeyboardInterrupt:
                         raise
+                    except unittest.SkipTest as e:
+                        result.addSkip(test, str(e))
                     except BaseException:
                         result.addError(
                             test,
--- a/src/zope/testrunner/tests/testrunner-debugging.rst
+++ b/src/zope/testrunner/tests/testrunner-debugging.rst
@@ -44,7 +44,7 @@ runner will enter pdb at that point:
     ...
     False
 
-Post-Mortem Dubigging
+Post-Mortem Debugging
 =====================
 
 You can also do post-mortem debugging, using the --post-mortem (-D)
@@ -107,6 +107,23 @@ converted to errors and can be debugged the same way:
     (Pdb) p y
     2
     (Pdb) c
+    Tearing down left over layers:
+      Tear down zope.testrunner.layer.UnitTests in N.NNN seconds.
+    False
+
+
+Skipping tests with ``@unittest.skip`` decorator does not trigger the
+post-mortem debugger:
+
+    >>> sys.stdin = Input('q')
+    >>> sys.argv = ('test -ssample3 --tests-pattern ^sampletests_d$'
+    ...             ' -t skipped -D').split()
+    >>> try: testrunner.run_internal(defaults)
+    ... finally: sys.stdin = real_stdin
+    ... # doctest: +NORMALIZE_WHITESPACE +REPORT_NDIFF +ELLIPSIS
+    Running zope.testrunner.layer.UnitTests tests:
+      Set up zope.testrunner.layer.UnitTests in N.NNN seconds.
+      Ran 1 tests with 0 failures, 0 errors and 1 skipped in N.NNN seconds.
     Tearing down left over layers:
       Tear down zope.testrunner.layer.UnitTests in N.NNN seconds.
     False
--- a/src/zope/testrunner/tests/testrunner-ex/sample3/sampletests_d.py
+++ b/src/zope/testrunner/tests/testrunner-ex/sample3/sampletests_d.py
@@ -37,6 +37,10 @@ class TestSomething(unittest.TestCase):
         y = 2
         assert x == y
 
+    @unittest.skip("skipped test")
+    def test_skipped(self):
+        self.fail('test should have been skipped')
+
 
 def f():
     x = 1


perrinjerome:
I guess I have to wait for https://github.com/zopefoundation/zope.testrunner/pull/140

davisagli (approved):
Looks good to me, aside from needing the other PR to update the test environment. I ran into this once or twice and am grateful for the fix.

icemac:
@perrinjerome You could rebase to `master`, #140 is merged now.

perrinjerome:
@icemac thanks - and thanks a lot also for the work behind #140   . I rebased and the tests are passing.

mgedmin (approved):
Thank you very much for fixing this!

icemac:
@d-maurer Do you think this PR is mergable?

d-maurer (approved):

perrinjerome:
Thanks every one let's merge then.

icemac:
I just released https://pypi.org/project/zope.testrunner/5.6/ including this PR.

perrinjerome:
I found one problem that is fixed on python 3.11. When running Zope tests with `-D`  with python 3.8, they fail with an error like this:

```
Error in test test_module_traversal (Products.PageTemplates.tests.testHTMLTests.HTMLTests)
  File "/opt/slapgrid/d02ebd529f47acc392f6b079c09d3e4b/parts/python3/lib/python3.8/unittest/case.py", line 741, in debug
    getattr(self, self._testMethodName)()
  File "/srv/slapgrid/slappart15/srv/project/Zope/src/Products/PageTemplates/tests/testHTMLTests.py", line 266, in test_module_traversal
    t()
  File "/opt/slapgrid/d02ebd529f47acc392f6b079c09d3e4b/parts/python3/lib/python3.8/unittest/case.py", line 227, in __exit__
    self._raiseFailure("{} not raised".format(exc_name))
  File "/opt/slapgrid/d02ebd529f47acc392f6b079c09d3e4b/parts/python3/lib/python3.8/unittest/case.py", line 164, in _raiseFailure
    raise self.test_case.failureException(msg)
AssertionError: NotFound not raised
```

There's a skipped tests here:

https://github.com/zopefoundation/Zope/blob/26291e673d94d65987e5dd5defec72572893fa03/src/Products/PageTemplates/tests/testChameleonTalesExpressions.py#L25-L28

the test class have a setup which sets a custom security policy and a tear down which resets it, in https://github.com/zopefoundation/Zope/blob/26291e673d94d65987e5dd5defec72572893fa03/src/Products/PageTemplates/tests/testHTMLTests.py#L72-L91

On python3.11 https://github.com/python/cpython/blob/3.11/Lib/unittest/case.py#L681-L688 the setup is not executed for skipped tests with debug, but on 3.8 https://github.com/python/cpython/blob/3.8/Lib/unittest/case.py#L740-L745 the setup is executed, then the SkipTest exception is raised and the teardown is not executed.

I believe it's not really a problem because this problem occurs only in very specific conditions and it's anyway fixed on python3.11

d-maurer:
Jérome Perrin wrote at 2022-12-13 21:46 -0800:
> ...
>There's a skipped tests here:
>
>https://github.com/zopefoundation/Zope/blob/26291e673d94d65987e5dd5defec72572893fa03/src/Products/PageTemplates/tests/testChameleonTalesExpressions.py#L25-L28
>
>the test class have a setup which sets a custom security policy and a tear down which resets it, in https://github.com/zopefoundation/Zope/blob/26291e673d94d65987e5dd5defec72572893fa03/src/Products/PageTemplates/tests/testHTMLTests.py#L72-L91

I think it is wrong that `tearDown` can be skipped after `setUp`
has succeeded:
the documentation (--> "https://docs.python.org/3.8/library/unittest.html#test-cases")
promises that `tearDown` is always called even if the test method
has raised an exception -- unless `setUp` failed.

Potentially, it is a Python bug though - `unittest` has had several
problems, especially related to the `debug` method.

> I believe it's not really a problem because this problem occurs only in very specific conditions and it's anyway fixed on python3.11

I agree so long as the problem only arises with the `-D` option,
more precisely, so long as the github tests remain all green.

perrinjerome:
Thanks for the feedback, actually this was fixed by the PRs for https://github.com/python/cpython/issues/80855  ( I understand now your comment that this was a python bug  ). The fix was also backported in 3.9 and 3.10 so this problem is only for python <= 3.8 .  

The new problem with these changes, on python <= 3.8. is that when running with `-D`, if a test is skipped its `setUp` is called but not its `tearDown`.  
On python >= 3.9 `setUp` and `tearDown` are not called for skipped tests (as long as they are skipped with a decorator, if the test raises a `SkipTest` exception, it's a bit different). 
Before these changes, the test runner stopped at the skip test, so this problem did not really exist.

It does not look so important to me , because only developers run with `-D`, tests from github actions should not be impacted.

---

PS: 

> I think it is wrong that `tearDown` can be skipped after `setUp`

this made me think that maybe python needs a patch like this:

```patch
diff --git a/Lib/unittest/case.py b/Lib/unittest/case.py
index 5167c5f843..4155fc643a 100644
--- a/Lib/unittest/case.py
+++ b/Lib/unittest/case.py
@@ -688,11 +688,13 @@ def debug(self):
             raise SkipTest(skip_why)
 
         self._callSetUp()
-        self._callTestMethod(testMethod)
-        self._callTearDown()
-        while self._cleanups:
-            function, args, kwargs = self._cleanups.pop()
-            self._callCleanup(function, *args, **kwargs)
+        try:
+            self._callTestMethod(testMethod)
+        finally:
+            self._callTearDown()
+            while self._cleanups:
+                function, args, kwargs = self._cleanups.pop()
+                self._callCleanup(function, *args, **kwargs)
 
     def skipTest(self, reason):
         """Skip this test."""
```

this would be correct regarding "`tearDown` is always called after successful `setUp`, but practically this would make use cases like zopetestrunner's `-D` completely different (and pointless sometimes I guess), because the debugger session would happen after tear down.

PR #156, opened 2023-11-26
Merge collective.xmltestreport code here

### LINKED ISSUE #148: Generate XML reports
## FEATURE REQUEST

`collective.xmltestreport` is a package that wraps `zope.testrunner` that mostly adds an `--xml` option to the test runner and produces XML reports of the tests being run.

Unfortunately `collective.xmltestreport` is only usable via `zc.buildout` and it is hardly maintained, although it is stable.

Getting the XML report code merged here would be great as this would mean:

- one less dependency
- being able to have XML reports without having to depend on `zc.buildout` (again less dependencies)
- being able to easily use CI facilities that use such XML files like Jenkins, GitLab, I guess GitHub can use them, I haven't checked it but it probably can 🤷🏾

gforcada:
This is the first step to get `collective.xmltestreport` code merged here, as outlined in #148 😄 

## Goals

There are a few goals on why getting `collective.xmltestreport` code merged here makes sense:

- being able to drop `collective.xmltestreport` from the toolchain while keeping the functionality
- get the XML reports without having to rely on `zc.buildout` integration from `collective.xmltestreport`
- being able to extract the test report to be displayed, either in [Jenkins](https://plugins.jenkins.io/junit/), [GitHub Actions](https://github.com/marketplace/actions/publish-test-results) or [GitLab](https://docs.gitlab.com/ee/ci/testing/unit_test_reports.html)

**Specially the GitHub/GitLab integration is the key.** 🤖 

Being able to show the detailed list of test failures and quick links of what failed, makes developers/contributors live much easier than having to scroll through log files that have plenty of output that needs quite some skills to understand and parse.

## Status

The **existing** test suite does work and if one uses the new `--xml` option, you get a `testreports` folder full of XML files ✨ 💫 

I made the most minimal changes possible, to see:

- does it actually work without too much fiddling with the existing code?
- is the direction the right one?

The minimal differences with what's on `collective.xmltestreport` are:

- I dropped `six`
- I dropped the pretty XML formatter and use the one provided by Python 
- I dropped the `zc.buildout` integration, this is already provided by `zope.testrunner` itself

Fixes #148.


icemac:
I like the general direction. In addition to tests for the code, there should also be more doc strings explaining the classes and methods/functions and some documentation which will show up at https://zopetestrunner.readthedocs.io/.

Currently the subunit tests fail in interesting ways. I have no idea why, but it seems to be caused by the changes here, as they were succeeding the last time on `master` 19 h ago.

gforcada:
Update: tests are written, older python versions are fixed, but now Python 3.12 fails 😕 

Do we need to get #158 merged first? 🤔

src/zope/testrunner/formatter.py
```diff
@@ -1334,3 +1346,293 @@ def _emit_exists(self, test):
...
+        self.failures = 0
+        self.time = 0.0
+
+    @property
+    def tests(self):
+        return len(self.testCases)
+
+    @property
+    def successes(self):
+        return self.tests - (self.errors + self.failures)
```
icemac, line 1365:
```suggestion
        return self.tests - self.errors - self.failures
```

I think this is easier to grasp.

src/zope/testrunner/formatter.py
```diff
@@ -1334,3 +1346,293 @@ def _emit_exists(self, test):
...
+
+
+class TestCaseInfo(object):
```
icemac, line 1368: Inheriting from `object` is no longer necessary.

What about making this class (and `TestSuiteInfo`) a dataclass?

src/zope/testrunner/formatter.py
```diff
@@ -1334,3 +1346,293 @@ def _emit_exists(self, test):
...
+
+        if seconds:
+            suite.time += seconds
+
+    def writeXMLReports(self, properties={}):
+
+        timestamp = datetime.now().isoformat()
+        hostname = socket.gethostname()
+
+        workingDir = Path(os.getcwd())
```
icemac, line 1549:
```suggestion
        workingDir = Path.cwd()
```

src/zope/testrunner/formatter.py
```diff
@@ -1334,3 +1346,293 @@ def _emit_exists(self, test):
...
+    """Compute the test class name from the test object."""
+    return '{0}.{1}'.format(test.__module__, test.__class__.__name__, )
+
+
+def filename_to_suite_name_parts(filename):
+    # lop off whatever portion of the path we have in common
+    # with the current working directory; crude, but about as
+    # much as we can do :(
+    filenameParts = Path(filename).parts
+    cwdParts = Path(os.getcwd()).parts
```
icemac, line 1390:
```suggestion
    cwdParts = Path.cwd().parts
```

src/zope/testrunner/formatter.py
```diff
@@ -1334,3 +1346,293 @@ def _emit_exists(self, test):
...
+                        del tb
+
+                    errorNode.set('message', errorMessage.split('\n')[0])
+                    errorNode.set('type', str(excType))
+                    # We need to decode here (and maybe elsewhere),
+                    # because prettyXML will try to encode it in
+                    # _escape_cdata using 'us-ascii' as encoding,
+                    # which of course fails for anything that is not ascii.
+                    text = (errorMessage + '\n\n' + stackTrace)
+                    errorNode.text = text
```
icemac, line 1602: I see no decoding here outside the comment. Is it still valid?

src/zope/testrunner/tests/testrunner-xml-report.rst
```diff
@@ -0,0 +1,193 @@
...
+
+Remove the files:
+
+    >>> _ = [x.unlink() for x in reports_folder.iterdir()]
+    >>> _ = reports_folder.rmdir()
+
+Errors reported
+===============
+
+If errors/faiulres happen, they are also reported on the XML files:
```
icemac, line 92:
```suggestion
If errors/failures happen, they are also reported in the XML files:
```

src/zope/testrunner/tests/testrunner-xml-report.rst
```diff
@@ -0,0 +1,193 @@
...
+
+Let's look at the XML reports:
+
+    >>> reports_folder.exists()
+    True
+
+There are quite some reports:
+
+    >>> len([x for x in reports_folder.iterdir()])
+    106
```
icemac, line 132: Does this test break if someone adds a new test to `directory_with_tests`?

icemac (changes requested):
It looks quite good so far. I just read through the code. I did not try it out.

I have some suggestions and like to have a change log entry.

When doing changes to this PR please do not force push, so I can easily see whet you have changed after this commit. (I will squash merge the PR in the end, so it will be only one commit.)

icemac:
> Update: tests are written, older python versions are fixed, but now Python 3.12 fails 😕

🚀 

> Do we need to get #158 merged first? 🤔

As it is currently only a WIP, we might merge this PR first even though we do not see whether it runs on 3.12.

gforcada:
@icemac thanks for the plentiful of comments, I, on purpose, did not try to refactor too much the code, as to be able to easily compare it with the existing version, and thus not introducing subtle bugs.

But if we want to refactor the code, then sure, I can apply all your suggestions 👍🏾 

As for not force pushing, sure, I can try to remember, but I'm so used to that, that I might forget 😅 I very much like to have small step-by-step commits that once you are about to merge, they all make sense. But if anyway you are going to squash all of them, sure, I can keep committing on top always.

icemac:
> But if we want to refactor the code, then sure, I can apply all your suggestions 👍🏾

The tests show that the code is good, so now it can be refactored.

src/zope/testrunner/formatter.py
gforcada: Done!

src/zope/testrunner/options.py
```diff
@@ -296,6 +296,11 @@ def _regex_search(s):
 avoid corrupting the subunit stream.
 """)
 
+reporting.add_argument(
+    '--xml', action='store_true', dest='xmlOutput', help="""\
+If given, XML reports will be written to the current directory.
```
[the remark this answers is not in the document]
gforcada: Done!

It will be a backwards incompatible change with regards to what `collective.xmltestreport` was doing.

We should mention it on the README, so whoever migrates from `collective.xmltestreport` to `zope.testrunner` with the new `--xml` that they know it is not a boolean, but a path 📚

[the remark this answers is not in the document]
gforcada: I updated the documentation to mention the new option and added an entry on CHANGES as well.

src/zope/testrunner/tests/testrunner-xml-report.rst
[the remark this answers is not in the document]
gforcada: Done! 😄

gforcada:
@icemac all feedback has been applied but in one, as I was lazy 😅 

I had to rebase on top of current master to solve a conflict though 😕

src/zope/testrunner/tests/testrunner-xml-report.rst
icemac: Let's keep it as it is for now.

src/zope/testrunner/options.py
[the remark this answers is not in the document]
icemac: Thank you!

icemac (approved):

icemac:
@gforcada Thank you for working on this PR, I am planning to use it in my projects as it can help there.
@mauritsvanrees I think this PR can be merged, do you have any thoughts about it?

gforcada:
@icemac anything I can do to get this merged? 😄 we can wait for next year though, no rush 😅

dataflake:
I don't see a good reason to delay the merge further.

Issue #157, opened 2023-12-05
zope.testrunner fails one test with Python 3.12, succeeding with 3.11


doko42:
seen when testing with the python 3.12 branch, the test succeed with 3.11.7:

Failure in test /home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/
tests/testrunner-colors.rst
Failed doctest test for testrunner-colors.rst
  File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/tests/te
strunner-colors.rst", line 0

----------------------------------------------------------------------
File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/tests/testrunner-colors.rst", line 165, in testrunner-colors.rst
Failed example:
    _ = testrunner.run_internal(defaults_skip)
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib/python3.12/doctest.py", line 1359, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest testrunner-colors.rst[14]>", line 1, in <module>
        _ = testrunner.run_internal(defaults_skip)
      File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/__init__.py", line 55, in run_internal
        runner.run()
      File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/runner.py", line 191, in run
        self.run_tests()
      File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/runner.py", line 292, in run_tests
        self.ran += run_layer(self.options, layer_name, layer, tests,
      File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/runner.py", line 473, in run_layer
        return run_tests(options, tests, layer_name, failures, errors, skipped,
      File "/home/packages/zope/zope.testrunner/.pybuild/cpython3_3.12/build/zope/testrunner/runner.py", line 399, in run_tests
        test(result)
      File "/usr/lib/python3.12/unittest/case.py", line 692, in __call__
        return self.run(*args, **kwds)
      File "/usr/lib/python3.12/unittest/case.py", line 662, in run
        result.stopTest(self)
[...]

PR #158, opened 2023-12-06
WIP: Partial fix for Python 3.12

mgedmin:
Python 3.12 doesn't call `startTest()` for skipped tests, which caused AttributeErrors for things initialized in our `TestResult.startTest()` (and `OutputFormatter.start_test()`).

This is not a complete fix: there are still test failures because they expected "ran 1 tests, 1 skipped" and now get "ran 0 tests, 0 skipped". I don't know if we want to keep the old method of counting; it kind of makes sense to me not to count skipped tests as running ones.

I'm not sure how to fix that without adding sys.version_info >= (3, 12) checks everywhere.

See #157 and https://github.com/python/cpython/issues/106584.

### DIFF
--- a/src/zope/testrunner/formatter.py
+++ b/src/zope/testrunner/formatter.py
@@ -52,7 +52,7 @@ class OutputFormatter:
 
     def __init__(self, options):
         self.options = options
-        self.last_width = 0
+        self.test_width = self.last_width = 0
         self.compute_max_width()
 
     progress = property(lambda self: self.options.progress)
--- a/src/zope/testrunner/runner.py
+++ b/src/zope/testrunner/runner.py
@@ -1019,9 +1019,10 @@ class TestResult(unittest.TestResult):
         self.testTearDown()
         # Without clearing, cyclic garbage referenced by the test
         # would be reported in the following test.
-        test.__dict__.clear()
-        test.__dict__.update(self._test_state)
-        del self._test_state
+        if hasattr(self, '_test_state'):
+            test.__dict__.clear()
+            test.__dict__.update(self._test_state)
+            del self._test_state
         cycles = None
         if (uses_refcounts and self.options.gc_after_test and
                 self.options.verbose >= 4):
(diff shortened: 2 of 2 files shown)


src/zope/testrunner/formatter.py
```diff
@@ -52,7 +52,7 @@ class OutputFormatter:
 
     def __init__(self, options):
         self.options = options
-        self.last_width = 0
+        self.test_width = self.last_width = 0
         self.compute_max_width()
```
mgedmin, line 56: TBH I don't know if it's enough to initialize test_width here, or if it should also be reset to 0 in every `test_success()`/`test_error()`/`test_failure()`.

d-maurer:
> This is not a complete fix: there are still test failures because they expected "ran 1 tests, 1 skipped" and now get "ran 0 tests, 0 skipped". I don't know if we want to keep the old method of counting; it kind of makes sense to me not to count skipped tests as running ones.

Does it really no longer count skipped tests at all (or only skipped tests which started before they decided they want the get skipped)?
I agree that skipped tests do not need to be counted as "run" but having a count of all skipped tests (run or not) can be helpful:  skipped tests indicates that the individual run did not exercise all potential cases.

> I'm not sure how to fix that without adding sys.version_info >= (3, 12) checks everywhere.

I assume that failing doctests are the main problem. In this case, a normalizer might be used as a central point to handle the version differences.

mgedmin:
There's also a risk that by skipping `startTest()` there's a combination of `-v`/`-p` options and a sequence of test runs/failures/skips that could mess up the output because I got the `test_width`/`last_width` logic wrong in this PR.

(Which is probably still better than the status quo, where the test runner crashes the first time it encounters a skipped test.)

d-maurer:
Marius Gedminas wrote at 2023-12-19 03:23 -0800:
>There's also a risk that by skipping `startTest()` there's a combination of `-v`/`-p` options and a sequence of test runs/failures/skips that could mess up the output because I got the `test_width`/`last_width` logic wrong in this PR.

I got an alternative idea:
apparently, the (original) error happened in our `stopTest`.
This means, we already have a way to use our own `stopTest`. Likely,
we can have our own `startTest` in a similar way.
In this case, we could let our `startTest` set a marker "startTest run"
on the test instance and `stopTest` look for it, if not there, `stopTest`
could call `startTest` before its real work.

mgedmin:
#159 is a simpler and more complete fix for this, closing.

PR #164, opened 2024-02-09
Fix XML tests when running in distribution resp. separately.

### LINKED ISSUE #163: Tests for version 6.3 fail when run from inside egg/wheel
See https://github.com/zopefoundation/zopetoolkit/actions/runs/7828690859/job/21359091342?pr=74#step:6:572

icemac:
Fixes #163

The problem was that the doctest report files have different names when running only a single test: they contain a path in the tmp directory. So I chose a unittest.

Additionally I refactored the test setup to use pathlib all over the place and to store the results in a temporary directory to prevent problems when running tests in parallel.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@
 6.4 (unreleased)
 ================
 
+- Fix XML tests when running in distribution resp. separately.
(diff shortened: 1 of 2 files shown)


dataflake (approved):

gforcada (approved):

icemac:

PR #165, opened 2024-02-17
Implicit namespaces support

### LINKED ISSUE #160: Support implicit namespaces
### What I did:

Converted `megrok.strictrequire` to a package using a PEP 420 implicit namespace.
See https://github.com/zopefoundation/megrok.strictrequire/pull/8

### What I expect to happen:

Tests run like before.

### What actually happened:

No tests were run because directories without an `__init__.py` are not considered to contain Python code.
See https://github.com/zopefoundation/zope.testrunner/blob/4f9048acb7f54677e6e6890b8491e069d136a2d6/src/zope/testrunner/find.py#L295-L299

### What version of Python and Zope/Addons I am using:

- OS independent
- Python 3.7+

⚠️ This feature is required to solve https://github.com/zopefoundation/meta/issues/194

gforcada:
Closes #160

Allow `zope.testrunner` to find tests suites for packages that follow the PEP 420 implicit namespaces.

Note that this works on test packages, like https://github.com/zopefoundation/megrok.strictrequire/pull/8

Test suite needs to be adjusted, but at least we have a reference implementation that _seems_ to be working for packages already using implicit namespaces.

To test it with `megrok.strictrequire`:

```shell
git clone git@github.com:zopefoundation/zope.testrunner
cd zope.testrunner
git checkout implicit-namespaces
git clone git@github.com:zopefoundation/megrok.strictrequire
cd megrok.strictrequire
git checkout pep440
cd ..
python3.11 -m venv venv
. venv/bin/activate
pip install -e megrok.strictrequire[test]
pip install -e .
zope-testrunner --test-path megrok.strictrequire/src
```

Most of the test failures of `zope.testrunner`'s own test suite is due to, probably, the test runner finding too much tests. I will dig into it later, but I wanted to post the PR as soon as it was minimally working ✨

### DIFF
--- a/src/zope/testrunner/find.py
+++ b/src/zope/testrunner/find.py
@@ -292,11 +292,6 @@ def find_test_files_(options):
 
     for (p, package) in test_dirs(options, {}):
         for dirname, dirs, files in walk_with_symlinks(options, p):
-            if dirname != p and not contains_init_py(options, files):
-                # This is not a plausible test directory. Avoid descending
-                # further.
-                del dirs[:]
-                continue
             root2ext = {}
             dirs[:] = [d for d in dirs if identifier(d)]
             d = os.path.split(dirname)[1]
--- a/src/zope/testrunner/tests/testrunner-edge-cases.rst
+++ b/src/zope/testrunner/tests/testrunner-edge-cases.rst
@@ -419,8 +419,10 @@ Post-mortem debugging of a docfile doctest failure:
 
 Post-mortem debugging with triple verbosity
 
+    >>> sys.stdin = Input('p x\nc')
     >>> sys.argv = 'test --layer samplelayers.Layer1$ -vvv -D'.split()
-    >>> testrunner.run_internal(defaults)
+    >>> try: testrunner.run_internal(defaults)
+    ... finally: sys.stdin = real_stdin
     Running tests at level 1
     Running samplelayers.Layer1 tests:
       Set up samplelayers.Layer1 in 0.000 seconds.


gforcada:
Status update: all tests pass and the only two adjustments on the tests seem to be fine enough.

I could move those tests adjustments to a separate PR to validate that, indeed, they are not needed for this PR to pass, but stand on their own.

jensens (approved):
The only side effect is, it may take longer to discover tests, right? But I think this is fine. If someone complains, testrunner can be configured to search more specific.

dataflake:
> Test suite needs to be adjusted, but at least we have a reference implementation that _seems_ to be working for packages already using implicit namespaces.

What do you mean by "test suite needs to be adjusted"? Do you mean existing packages need their tests changed?

jensens:
> What do you mean by "test suite needs to be adjusted"? Do you mean existing packages need their tests changed?

Since it did not traverse into non-`__init__.py`-containing directories and now does, it may result in longer discovery runs. To avoid this we may want to configure the zope.testrunner to search more precisely using other `--path` and/or `--test-pattern` configurations.
OTOH I do not expect it to discover tests, which it had not found before - so, it won't break.

gforcada:
@dataflake sorry, that was an interim comment, when I first pushed this PR, I knew that the code was working, as I used the `megrok.staticrequire` distribution as a test sample, but I was not sure about `zope.testrunner`'s own test suite.

I was expecting more breakage, but turned out not to be 👍🏾 

Regarding test discovery, I see a problem, like we faced in `z3c.dependencychecker` when we added support for implicit namespaces (https://github.com/reinout/z3c.dependencychecker/pull/126) and the two follow up PRs to improve the time it takes to run: https://github.com/reinout/z3c.dependencychecker/pull/128

Basically, ignore folders with leading dots, `__pycache__` or `node_modules` for example.

dataflake:
> Basically, ignore folders with leading dots, `__pycache__` or `node_modules` for example.

It looks like the current code here only omits leading dot folders with the regular expression `identifier`:

https://github.com/zopefoundation/zope.testrunner/blob/a68a8669afee2e0248082a5ce40232a2249e5aaf/src/zope/testrunner/find.py#L28

This expression is used to filter out folders while walking the tree:

https://github.com/zopefoundation/zope.testrunner/blob/a68a8669afee2e0248082a5ce40232a2249e5aaf/src/zope/testrunner/find.py#L301

I suggest adding at least `__pycache__` to the undesirable folder names there.

davisagli:
I'm not sure what would be the simplest way to implement this, but it would be nice if the test discovery skipped any paths that are matched by `.gitignore` configuration.

icemac:
> I'm not sure what would be the simplest way to implement this, but it would be nice if the test discovery skipped any paths that are matched by `.gitignore` configuration.

There is https://github.com/thehanimo/py-gitignore under MIT license which seems to be capable of checking whether a path is ignored via `.gitignore`.

d-maurer:
Michael Howitz wrote at 2024-2-20 23:24 -0800:
>> I'm not sure what would be the simplest way to implement this, but it would be nice if the test discovery skipped any paths that are matched by `.gitignore` configuration.
>
>There is https://github.com/thehanimo/py-gitignore under MIT license which seems to be capable of checking whether a path is ignored via `.gitignore`.

Not sure whether depending on `.gitignore` is a good approach:
it would mean that test runs behave differently between
run from a git checkout (with .gitignore) and from e.g. an
installation (without ".gitignore")
(and maybe even a source distribution).

When I have understood this correctly, we run tests
not from a git checkout but from a wheel for some projects.

dataflake:
I agree with Dieter and I don't think there's much gain for the price of yet another dependency or manually written code that needs maintaining going forward. My suggestion still stands, expand the current exclusion of dot folders by adding `__pycache__` and get on with this.

davisagli:
@d-maurer Good point, I agree that's a compelling reason to not use `.gitignore`.

gforcada:
I adjusted the code just slightly to ignore well known folders we want to skip.

I added some minimal tests to ensure we know what we are accepting/rejecting right now.

🍀

jensens:
As an idea for a future iteration as a new PR (to get this one here merged soon): Get the `IGNORE_FOLDERS` configuration from a custom section in `pyproject.toml`?

dataflake (approved):

jensens (approved):

gforcada:
@jensens yes, I thought about that as well, that one way or another users could provide their own list of ignored folders 👍🏾 

But yes, let's merge this, a new release, and see how much it breaks 🍀 😅

dataflake:
Release 6.4 is now published

PR #172, opened 2024-06-01
use more `...` in doctests to make them pass for Python 3.13

### LINKED ISSUE #167: Tests are broken on Python 3.13a6+
See https://github.com/zopefoundation/zope.testrunner/actions/runs/8772493667/job/24071512954

It looks like some changes in GC. And the doctests look like a pain to fix.

d-maurer:
Fixes #167.

Sadly, `zope.testrunner` (mostly) uses doctests as its tests. Complex doctests tend to be very vulnerable regarding changes between Python versions.

This PR agressively uses `...` in doctests to make them pass for Python 3.13 and earlier Python versions. At some places, the goal might be achieved with less drastic steps (especially regarding the coloring tests).

If you run the tests locally, you may still observe some test failures if a `.pdbrc` file is effective (see "https://github.com/python/cpython/issues/119824").

There is no `CHANGES.rst` entry because this PR is part of a more general `support Python 3.13` one.


dataflake (approved):
I share your frustration with doctests. They add nothing but make it impossible to debug issues.

tseaver:
@dataflake The original goal of doctests was to ensure that examples used in real documentation didn't bitrot.  I'm still in favor of that goal, but would support any effort to replace doctests-used-for-testing:  those tend to suck both as cos and as tests.

One clue the their existence is that the tests are using `.txt` files which are *not* part of the Sphinx docs (and no, I don't want them included blindly, as they are mostly terrible as docs).

PR #176, opened 2024-07-31
Remove setuptools fossils.

mauritsvanrees:


mauritsvanrees:
Some actual fixes are needed here.  The [`testrunner-eggsupport.rst`](https://github.com/zopefoundation/zope.testrunner/actions/runs/10178731246/job/28152946032?pr=176#step:6:34) tests fail:

```
File "/home/runner/work/zope.testrunner/zope.testrunner/src/zope/testrunner/tests/testrunner-eggsupport.rst", line 40, in testrunner-eggsupport.rst
Failed example:
    from zope.testrunner.eggsupport import ftest
Exception raised:
      File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/doctest.py", line 1368, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest testrunner-eggsupport.rst[15]>", line 1, in <module>
        from zope.testrunner.eggsupport import ftest
      File "/home/runner/work/zope.testrunner/zope.testrunner/src/zope/testrunner/eggsupport.py", line 4, in <module>
        from setuptools.command.test import ScanningLoader
    ImportError: cannot import name 'ScanningLoader' from 'setuptools.command.test' (/home/runner/work/zope.testrunner/zope.testrunner/.tox/py312/lib/python3.12/site-packages/setuptools/command/test.py)
```

They only fail on Python 3.12 and 3.13.  These get `setuptools==72.1.0`.  Python 3.11 gets `setuptools==70.1.0` so there it still works.

The [`eggsupport.py` file](https://github.com/zopefoundation/zope.testrunner/blob/6.4/src/zope/testrunner/eggsupport.py) starts with this, which for obvious reasons is not available anymore on setuptools 72:

```
from setuptools.command.test import ScanningLoader
from setuptools.command.test import test as BaseCommand
```

I wonder if our `eggsupport` module is by definition broken and unfixable in setuptools 72+ , and if we should simply make a note of this in the changelog and ignore the test file on that setuptools version.

mauritsvanrees:
I did not try it, but it looks like this module is only used when you call `python setup.py ftest`, which is a distutils command we define.  I have never used this.  Since the standard `python setup.py test` should no longer be used and has been removed, it is probably fine if our `ftest` command is broken on setuptools 72+ as well.  We should add this observation to the changelog.

Apart from making the tests pass, we could either do nothing, or catch the ImportError and provide an own error message.

gforcada:
Sounds like a good opportunity to clean some code 👍🏾

mauritsvanrees:
I have deprecated our `ftest` command and scheduled it for removal in `zope.testrunner` 7.

Interestingly we have a [hack in our own `setup.py`](https://github.com/zopefoundation/zope.testrunner/blob/6.4/setup.py#L175-L177) that replaces the standard `test` command with a [`custom_test` command](https://github.com/zopefoundation/zope.testrunner/blob/6.4/setup.py#L78).  So this `custom_test` command is used when you run `python setup.py test` on the `setup.py` of this same package.

Ah, and this works (sort of) because we do [`from setuptools.command.test import test`](https://github.com/zopefoundation/zope.testrunner/blob/6.4/setup.py#L23) at the top of our `setup.py`.  So that won't work either once the upstream `test` command is completely gone.

It is probably best to remove that code now.  We probably do not need to support running `python setup.py test` (or `python setup.py ftest`) on ourselves.

icemac:
I am fine with not having a deprecation period but upgrading the major version number.

`python setup.py ftest` or `python setup.py test` is no longer/should be no longer used in our own tests. (Writing this, I remember that `zope.interface` is maybe still using it.)

mauritsvanrees:
Part of this was done by @d-maurer in PR #178.
I will close the current PR and make a new one with the remaining needed changes.

PR #179, opened 2024-08-23
Modernize setuptools usage & exclude tests from wheel

carlsmedstad:
Hey :wave:

I maintain the [Arch Linux package for zope.testrunner](https://archlinux.org/packages/extra/any/python-zope-testrunner/). This is a set of packaging changes that, in my view, are appropriate - the primary one being to stop including tests in the distributed wheel.

As the wheel's primary purpose is to be installed in a virtualenv or on a system, the tests for testrunner itself are unnecessary to include. Including them also makes distro-tooling such as Arch Linux's namcap produce noisy
warnings regarding missing dependencies.

Before these changes:

```sh
$ rm -rf build dist/ src/zope.testrunner.egg-info/; python -m build --wheel
...
Successfully built zope.testrunner-6.6.dev0-py3-none-any.whl
$ find build/lib/zope/testrunner/ -iname tests
build/lib/zope/testrunner/tests
```

After:

```sh
$ rm -rf build dist/ src/zope.testrunner.egg-info/; python -m build --wheel
...
Successfully built zope.testrunner-6.6.dev0-py3-none-any.whl
$ find build/lib/zope/testrunner/ -iname tests
```

The tests are still included in the sdist as per the MANIFEST.in file:

```sh
$ rm -rf build dist/ src/zope.testrunner.egg-info/; python -m build --sdist
...
Successfully built zope_testrunner-6.6.dev0.tar.gz
$ tar -xf dist/zope_testrunner-6.6.dev0.tar.gz
$ find zope_testrunner-6.6.dev0/ -iname tests
zope_testrunner-6.6.dev0/src/zope/testrunner/tests
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -91,7 +91,7 @@ setup(
     long_description=long_description,
     author='Zope Foundation and Contributors',
     author_email='zope-dev@zope.dev',
-    packages=find_namespace_packages(where='src'),
+    packages=find_namespace_packages(where='src', exclude=['zope.testrunner.tests*']),
     package_dir={'': 'src'},
     classifiers=[
         "Development Status :: 5 - Production/Stable",
@@ -124,6 +124,6 @@ setup(
         'distutils.commands': [
             'ftest = zope.testrunner.eggsupport:ftest'],
     },
-    include_package_data=True,
+    include_package_data=False,
     zip_safe=False,
 )
--- a/setup.py
+++ b/setup.py
@@ -60,11 +60,6 @@ else:
 import os
 os.chdir(%r)
 
-# The following unused imports are dark magic that makes the tests pass on
-# Python 3.5 on Travis CI.  I do not understand why.
-import zope.exceptions.exceptionformatter
-import zope.testing
-
 import zope.testrunner
 if __name__ == '__main__':
     zope.testrunner.run([
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@
 ##############################################################################
 import os
 
-from setuptools import find_namespace_packages
+from setuptools import find_packages
 from setuptools import setup
 
 
@@ -91,7 +91,7 @@ setup(
     long_description=long_description,
     author='Zope Foundation and Contributors',
     author_email='zope-dev@zope.dev',
-    packages=find_namespace_packages(where='src', exclude=['zope.testrunner.tests*']),
+    packages=find_packages(where='src', exclude=['zope.testrunner.tests*']),
     package_dir={'': 'src'},
     classifiers=[
         "Development Status :: 5 - Production/Stable",
@@ -114,6 +114,7 @@ setup(
         "Topic :: Software Development :: Libraries :: Python Modules",
         "Topic :: Software Development :: Testing",
     ],
+    namespace_packages=['zope'],
     python_requires='>=3.7',
     install_requires=INSTALL_REQUIRES,
     tests_require=TESTS_REQUIRE,
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@
 ##############################################################################
 import os
 
+from setuptools import find_namespace_packages
 from setuptools import setup
 
 
@@ -90,10 +91,7 @@ setup(
     long_description=long_description,
     author='Zope Foundation and Contributors',
     author_email='zope-dev@zope.dev',
-    packages=[
-        "zope",
-        "zope.testrunner",
-    ],
+    packages=find_namespace_packages(where='src'),
     package_dir={'': 'src'},
     classifiers=[
         "Development Status :: 5 - Production/Stable",
@@ -116,7 +114,6 @@ setup(
         "Topic :: Software Development :: Libraries :: Python Modules",
         "Topic :: Software Development :: Testing",
     ],
-    namespace_packages=['zope'],
     python_requires='>=3.7',
     install_requires=INSTALL_REQUIRES,
     tests_require=TESTS_REQUIRE,
--- a/src/zope/__init__.py
+++ b/src/zope/__init__.py
@@ -1,1 +1,0 @@
-__import__('pkg_resources').declare_namespace(__name__)
--- a/src/zope/__init__.py
+++ b/src/zope/__init__.py
@@ -1,0 +1,1 @@
+__import__('pkg_resources').declare_namespace(__name__)


carlsmedstad:
One additional reason to exclude the tests from the wheel is that they contain invalid Python, and as such make the wheel fail installation without `--no-compile`:

```sh
$ rm -rf build dist/ src/zope.testrunner.egg-info/; python -m build --wheel
...
Successfully built zope.testrunner-6.6.dev0-py3-none-any.whl
$ python -m installer --destdir=tmp_install dist/*.whl
*** Error compiling 'tmp_install/usr/lib/python3.12/site-packages/zope/testrunner/tests/testrunner-ex/sample2/badsyntax.py'...
  File "/usr/lib/python3.12/site-packages/zope/testrunner/tests/testrunner-ex/sample2/badsyntax.py", line 16
    importx unittest  # noqa: E999
            ^^^^^^^^
SyntaxError: invalid syntax

*** Error compiling 'tmp_install/usr/lib/python3.12/site-packages/zope/testrunner/tests/testrunner-ex/sample2/badsyntax.py'...
  File "/usr/lib/python3.12/site-packages/zope/testrunner/tests/testrunner-ex/sample2/badsyntax.py", line 16
    importx unittest  # noqa: E999
            ^^^^^^^^
SyntaxError: invalid syntax
```

setup.py
```diff
@@ -121,7 +113,6 @@ def read(*names):
         "Topic :: Software Development :: Libraries :: Python Modules",
         "Topic :: Software Development :: Testing",
     ],
-    namespace_packages=['zope'],
```
icemac, line 124: This change goes too far by switching to implicit namespaces (PEP-420) – It would require all other 117 packages of this namespace to switch over at the same time.
See https://github.com/zopefoundation/meta/issues/194 where we track our progress regarding this PEP.

If the issue cannot be solved without the switch to PEP-420, it has to wait until we have everything in place to to the swtich for all packages.

icemac (changes requested):

icemac:
Thank you for your PR. I like the idea to get rid of these warnings during installation.

d-maurer:
Carl Smedstad wrote at 2024-8-22 22:59 -0700:
> ...
>As the wheel&#39;s primary purpose is to be installed in a virtualenv or on a system, the tests for testrunner itself are unnecessary to include.

I like to have the tests around for `zopefoundation` packages:
the tests often can be used as additional documentation
(especially for situations sufficuently rare not to be covered
by the "standard" documentation).

In my view, including the tests in the distributions is only
a marginal overhead.
Of course, I am not familiar with distro-tools.

setup.py
carlsmedstad: > This change goes too far by switching to implicit namespaces (PEP-420) – It would require all other 117 packages of this namespace to switch over at the same time.

I have to admit this is my first time interacting with namespace packages, so I might be out on a limb here, but this does not seem to be the case:

```sh
$ python -m venv .venv && source ./.venv/bin/activate
(.venv) $ pip install zope.interface
...
Successfully installed setuptools-73.0.1 zope.interface-7.0.1
(.venv) $ pip install dist/zope.testrunner-6.6.dev0-py3-none-any.whl
...
Successfully installed zope.exceptions-5.1 zope.testrunner-6.6.dev0
(.venv) $ python -c "import zope.interface; import zope.testrunner;"
```

I.e. testrunner built from this branch can be installed and imported without issues alongside `zope.interface`, which uses an explicit namespace declaration. Maybe I'm missing something?

icemac: It's a complicated matter: It sometimes depends on import order and installing packages using `zc.buildout` (as done often in the Zope community) also seems to play a role. So our plans are to do the switch for all packages together to circumvent the problems … but this will require time and coordination effort.

carlsmedstad: Aha, I see. I've removed the migration to implicitly declared namespace in https://github.com/zopefoundation/zope.testrunner/pull/179/commits/e9b96914fbabace1b66b842e7b0fb13695026b7b. This leaves the exclusion of the tests from the wheel, is that something you could consider merging?

icemac:
I'd give it a go, to try out wheels without tests. @dataflake What do you think?

dataflake:
I don't care either way. But there's a failing test that did not fail before.

carlsmedstad:
I don't have the possibility of testing locally on Windows, and if I understand correctly the CI is not automatically triggered on push for me, so I'm afraid this makes it difficult for me to iterate on a Windows+PyPy test failure.

With that said, could this failure be timing-related? It only occurring on one platform suggests that. The sleep at [src/zope/testrunner/tests/test_threadsupport.py#L121](https://github.com/zopefoundation/zope.testrunner/blob/master/src/zope/testrunner/tests/test_threadsupport.py#L121) looks like a candidate for the cause to me.

mauritsvanrees:
I see no failing tests anymore.  Maybe someone reran the failing test and it was solved.

In general in most packages, I very much want to keep the tests in both source distribution and binary wheel.  This is because in Plone we run the tests of about 100 packages combined.  We are not going to use editable installs for all those packages, but we install the actual distributions from PyPI (plus a few editable installs).  And we use `zope.testrunner` for running the tests.
But in Plone we don't need to run the `zope.testrunner` tests, so from a Plone perspective I don't mind.

So I am -0 on this change: I prefer to keep the tests, but I don't mind much.

carlsmedstad:
> I see no failing tests anymore. Maybe someone reran the failing test and it was solved.
> 
> In general in most packages, I very much want to keep the tests in both source distribution and binary wheel. This is because in Plone we run the tests of about 100 packages combined. We are not going to use editable installs for all those packages, but we install the actual distributions from PyPI (plus a few editable installs). And we use `zope.testrunner` for running the tests. But in Plone we don't need to run the `zope.testrunner` tests, so from a Plone perspective I don't mind.
> 
> So I am -0 on this change: I prefer to keep the tests, but I don't mind much.

I appreciate your use-case. But, while distributing tests in wheels is (still) somewhat prevalent, in my experience it is definitely a minority of projects doing so, so I would avoid relying on it.

While slightly less convenient for your use-case, but still very viable, you could download and install the sdist from PyPI:

```sh
$ pip download --no-binary :all: zope.testrunner
Collecting zope.testrunner
  Downloading zope.testrunner-6.5.tar.gz (155 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting setuptools (from zope.testrunner)
  Using cached setuptools-74.0.0.tar.gz (1.4 MB)
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting zope.exceptions (from zope.testrunner)
  Downloading zope.exceptions-5.1.tar.gz (31 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting zope.interface (from zope.testrunner)
  Downloading zope.interface-7.0.3.tar.gz (252 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Saved ./zope.testrunner-6.5.tar.gz
Saved ./setuptools-74.0.0.tar.gz
Saved ./zope.exceptions-5.1.tar.gz
Saved ./zope.interface-7.0.3.tar.gz
Successfully downloaded zope.testrunner setuptools zope.exceptions zope.interface
$ pip install *.tar.gz
Processing ./setuptools-74.0.0.tar.gz
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Processing ./zope.exceptions-5.1.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Processing ./zope.interface-7.0.3.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Processing ./zope.testrunner-6.5.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: setuptools, zope.exceptions, zope.interface, zope.testrunner
  Building wheel for setuptools (pyproject.toml) ... done
  Created wheel for setuptools: filename=setuptools-74.0.0-py3-none-any.whl size=1301729 sha256=dd7693ef28f3a714c724af409c1ddf80612caa7179840f896d462983a574b9c8
  Stored in directory: /home/carsme/.cache/pip/wheels/9c/5d/c7/01881acca67b97114f769d8a0d677744f8eb2aad907d1a208d
  Building wheel for zope.exceptions (pyproject.toml) ... done
  Created wheel for zope.exceptions: filename=zope.exceptions-5.1-py3-none-any.whl size=19884 sha256=794e3df5dba6bee9c4a20351ab1c21447b50f5c9efa13ff392fa47a7c625b473
  Stored in directory: /home/carsme/.cache/pip/wheels/8b/93/8d/a8a6204e91ea82ec83d411bee32f85f1b9bc9626b791467f2d
  Building wheel for zope.interface (pyproject.toml) ... done
  Created wheel for zope.interface: filename=zope.interface-7.0.3-cp312-cp312-linux_x86_64.whl size=259581 sha256=0ceca0d1d0a3cd58dc63b6ad4f6f0b92a0ae102ed72bba79aafd653db637b675
  Stored in directory: /home/carsme/.cache/pip/wheels/9f/3c/81/c8234495196a05e2de81195c6f541159e2e0967a910c2c76da
  Building wheel for zope.testrunner (pyproject.toml) ... done
  Created wheel for zope.testrunner: filename=zope.testrunner-6.5-py3-none-any.whl size=233733 sha256=c0e216b580dd6399754b706736498a453b232f6e62749db9d516f5dffe16368c
  Stored in directory: /home/carsme/.cache/pip/wheels/80/28/a7/a37919aa2a91c9f17eb3d1f95829ed316d94073702de10ee0e
Successfully built setuptools zope.exceptions zope.interface zope.testrunner
Installing collected packages: setuptools, zope.interface, zope.exceptions, zope.testrunner
Successfully installed setuptools-74.0.0 zope.exceptions-5.1 zope.interface-7.0.3 zope.testrunner-6.5
```

dataflake:
> > So I am -0 on this change: I prefer to keep the tests, but I don't mind much.
> 
> I appreciate your use-case. But, while distributing tests in wheels is (still) somewhat prevalent, in my experience it is definitely a minority of projects doing so, so I would avoid relying on it.

This argument doesn't work here. What @mauritsvanrees means is testing projects that are fully under "our" control, the Plone and Zope folks. We're talking about several hundreds of projects that are set up similarly and almost all of them do ship the tests in the wheel. What you're proposing here is to create a one-off in this specific project that will behave differently. As one of the Zope release managers and a co-maintainer of the toolchain and the methodologies that are used across hundreds of repositories I am very sceptical of one-offs that behave differently from the rest.

mauritsvanrees:
I am actually surprised that the source distribution contains different files than the wheel.  I mean sure, there are different metadata files, and files may be in a different place, and a file like `setup.py` won't be in the wheel, but I would expect everything within the `src` directory to be contained in both.  I am used to excluding files with `MANIFEST.in`, and that has the same effect on source dist and wheel.  But apparently the way you do it within `setup.py` has effect on the wheel, but not on the source dist.

The good thing about this is that, as you point out, the tests are still available when you install the source dist, so that is available as fallback.  The bad thing is that it matters whether you get an install of the source dist or of the wheel.  If you run the tests of 10 packages combined, each with 10 tests, then it could be that on one machine you see 100 tests being run and on another 90.  Good luck debugging that.

So it may be best if you can patch these files out on your end.

dataflake:
After this discussion I am against merging this PR.

d-maurer:
Jens Vagelpohl wrote at 2024-9-1 00:10 -0700:
>After this discussion I am against merging this PR.

+1

icemac:
I see consensus in rejecting this PR.
Thank you to everyone who joined the discussion and let us come to a conclusion.

PR #183, opened 2024-10-11
Drop support for Python 3.7 + officially support Python 3.13

icemac:


icemac:
@d-maurer We seem to work against each other here. I have no Windows machine at hand, so I have to use GHA to try to fix the issues.

d-maurer:
> @d-maurer We seem to work against each other here. I have no Windows machine at hand, so I have to use GHA to try to fix the issues.

@icemac Sorry, I step back.

Locally, I added the "--Return--" checker rules to the `win32` checker (which has been missing).

dataflake (approved):

d-maurer (approved):

icemac:

PR #187, opened 2024-12-09
make signatures in `tb_format` compatible with Python 3.12+

### LINKED ISSUE #186: Since Python 3.12, tbformat.py is broken
The signature of `traceback.format_exception()` changed in Python 3.12. testrunner replaces it with a different signature, breaking other places within `traceback`.

https://github.com/zopefoundation/zope.testrunner/blob/a7b36ab8e532666c39d83b70a3fd03e324e61822/src/zope/testrunner/tb_format.py#L60

for Python >= 3.12 should match:

`def format_exception(exc, /, value=_sentinel, tb=_sentinel, limit=None, \
                     chain=True, **kwargs):`

`print_exception` needs similar changes.

d-maurer:
Fixes #186.

`zope.testrunner` overrides the `traceback` funktions `print_exception` and `format_exception` for two reasons:
1. Python's functions have `True` as default for the parameter `chain`  causing a nested traceback display
2. Python's functions mark some traceback lines with `^^^^^^`.
Both risk to let existing doctests fail.

Python 3.12 has changed the signatures of its functions above. This PR modifies the signature of its overriding functions to be compatible with all supported Python versions.

### DIFF
--- a/src/zope/testrunner/tb_format.py
+++ b/src/zope/testrunner/tb_format.py
@@ -22,6 +22,13 @@ import zope.exceptions.exceptionformatter
 import zope.testrunner.feature
 
 
+try:
+    from traceback import _sentinel
+except ImportError:
+    # before 3.12
+    _sentinel = False
+
+
 def _iter_chain(exc, custom_tb=None, seen=None):
     if seen is None:
         seen = set()
@@ -47,9 +54,9 @@ def _parse_value_tb(exc, value, tb):
     # Taken straight from the traceback module code on Python 3.10, which
     # introduced the ability to call print_exception with an exception
     # instance as first parameter
-    if (value is None) != (tb is None):
+    if (value is _sentinel) != (tb is _sentinel):
         raise ValueError("Both or neither of value and tb must be given")
-    if value is tb is None:
+    if value is tb is _sentinel:
         if exc is not None:
             return exc, exc.__traceback__
         else:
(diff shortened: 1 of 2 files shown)


dataflake (approved):

src/zope/testrunner/tb_format.py
```diff
@@ -22,6 +22,13 @@
 import zope.testrunner.feature
 
 
+try:
+    from traceback import _sentinel
+except ImportError:
+    # before 3.12
+    _sentinel = False
```
davisagli, line 29: I would do 
```suggestion
    _sentinel = object()
```
so it can't be accidentally confused with passing a value of `False`, but I guess it probably doesn't matter.

d-maurer:
David Glick wrote at 2024-12-9 16:10 -0800:
>@davisagli commented on this pull request.
>
>> @@ -22,6 +22,13 @@
> import zope.testrunner.feature
>
>
>+try:
>+    from traceback import _sentinel
>+except ImportError:
>+    # before 3.12
>+    _sentinel = False
>
>I would do
>```suggestion
>    _sentinel = object()
>```

Thank you for your comment!
It made me look again carefully at the code.

In fact, `_sentinel` is defined from Python 3.10 on (not 3.12).
Before, the functions required that all of `etype`, `value`, `tb`
are passed: `value` could have a false value, `tb` could be `None`
to indicate unavailability.
Thus, `_sentinel = False` is wrong for Python before 3.10;
it must be `_sentinel = None`.

d-maurer:
@davisagli Turns out you have been right (and I was wrong): as `_sentinel` value, I must use something **not** used by the application.

PR #190, opened 2025-02-04
Add option `--only-level=level` to run tests only at the specified level.

### LINKED ISSUE #188: Add an option to run only tests on a specific level
## FEATURE REQUEST

There is currently the option `--at-level` which runs the tests at the given level and below.
For our integration tests it would be helpful to be able to run only the tests on a specific level (nothing below or above).

### Implementation idea

Add a new option `--only-at-level` which implements the requested behaviour. (See #139 for inspiration where to put it in the code.)
This would not interfere with any existing options and behavior.

icemac:
Fixes #188.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,8 @@
 6.7 (unreleased)
 ================
 
-- Nothing changed yet.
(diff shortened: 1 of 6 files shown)


dataflake (approved):

icemac:

PR #198, opened 2025-05-27
Add ``--auto-path`` option.

mauritsvanrees:
This automatically adds the path of all packages from the ``--package`` option to the test paths.

Use case: install all Plone packages with pip, some with an editable install, install `zope.testrunner`, and then call for example `zope-testrunner --auto-path -s plone.memoize -s plone.base` without having to set the `--test-path` for each package.

Otherwise this gets tricky when some packages are in site-packages and some editable in a `src` directory, and for editable installs their code may or may not be within a `src` sub directory:

```
zope-testrunner --test-path=src/plone.base/src --test-path=.venv/lib/python3.13/site-packages -s plone.base -s plone.memoize
```

With Buildout you can easily do it like this:

```
[test]
recipe = zc.recipe.testrunner
eggs = ${buildout:test-eggs}
```

With the `--auto-path` option you say: "I don't care where it exactly is, just find the tests of that package and run them."  You can still pass other `--test-path` options if you need.

I am not sure how to add tests for this feature.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,8 @@
 7.4 (unreleased)
 ================
 
-- Nothing changed yet.
+- Add ``--auto-path`` option.
+  This automatically adds the path of all packages from the ``--package`` option to the test paths.
 
 
 7.3 (2025-05-15)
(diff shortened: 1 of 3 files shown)


dataflake (approved):
Sounds like a nice feature

PR #205, opened 2026-03-31
Skip _SubTest objects

frenzymadness:
We are building Python packages in Fedora Linux with Python 3.15 already and one test failed with the following error:

```
Failure in test test_xml_report_with_errors (zope.testrunner.tests.test_xml_output.TextXMLOutputWithErrors.test_xml_report_with_errors)
  File "/usr/lib64/python3.15/unittest/case.py", line 58, in testPartExecutor
    yield
  File "/usr/lib64/python3.15/unittest/case.py", line 667, in run
    self._callTestMethod(testMethod)
  File "/usr/lib64/python3.15/unittest/case.py", line 613, in _callTestMethod
    result = method()
  File "/builddir/build/BUILD/python-zope-testrunner-8.2-build/BUILDROOT/usr/lib/python3.15/site-packages/zope/testrunner/tests/test_xml_output.py", line 133, in test_xml_report_with_errors
    self.assertEqual(len([x for x in self.reports_folder.iterdir()]), 106)
  File "/usr/lib64/python3.15/unittest/case.py", line 925, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib64/python3.15/unittest/case.py", line 918, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 107 != 106
```

Excluding _SubTest objects makes the test pass because the number of files matches the expected one. I'm not sure why this behavior is different in CPython 3.15 as I wasn't able to find the respective change there.

### DIFF
--- a/src/zope/testrunner/tests/test_xml_output.py
+++ b/src/zope/testrunner/tests/test_xml_output.py
@@ -133,7 +133,8 @@ class TextXMLOutputWithErrors(Base):
         # Python 3.15+ includes an additional unittest.case._SubTest.xml file
         # containing subtest information
         expected_count = 107 if sys.version_info >= (3, 15) else 106
-        self.assertEqual(len([x for x in self.reports_folder.iterdir()]), expected_count)
+        self.assertEqual(
+            len([x for x in self.reports_folder.iterdir()]), expected_count)
 
     def test_xml_report_with_errors_details(self):
         sys.argv = 'test --tests-pattern ^sampletests(f|_e|_f)?$ '.split()
--- a/src/zope/testrunner/tests/test_xml_output.py
+++ b/src/zope/testrunner/tests/test_xml_output.py
@@ -24,6 +24,8 @@ from pathlib import Path
 from zope import testrunner
 
 
+PY314_OR_OLDER = sys.version_info < (3, 15)
+
 class Base(unittest.TestCase):
 
     def tearDown(self):
@@ -130,9 +132,9 @@ class TextXMLOutputWithErrors(Base):
         self._run_tests()
 
         self.assertTrue(self.reports_folder.exists())
-        # Python 3.15+ includes an additional unittest.case._SubTest.xml file
+        # Python < 3.15 does not include a unittest.case._SubTest.xml file
         # containing subtest information
-        expected_count = 107 if sys.version_info >= (3, 15) else 106
+        expected_count = 106 if PY314_OR_OLDER else 107
         self.assertEqual(
             len([x for x in self.reports_folder.iterdir()]), expected_count)
 
--- a/src/zope/testrunner/tests/test_xml_output.py
+++ b/src/zope/testrunner/tests/test_xml_output.py
@@ -26,6 +26,7 @@ from zope import testrunner
 
 PY314_OR_OLDER = sys.version_info < (3, 15)
 
+
 class Base(unittest.TestCase):
 
     def tearDown(self):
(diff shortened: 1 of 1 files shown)


gforcada:
Using

```shell
uvx --from tox-uv tox run -e py315 -- -t test_xml_report_with_errors
```

With a `traceback` on the test failure:

```python
        self.assertTrue(self.reports_folder.exists())
        breakpoint()
        self.assertEqual(len([x for x in self.reports_folder.iterdir()]), 106)
```

Indeed, there is one extra file on `self.reports_folder`: `unittest.case._SubTest.xml`

Reading that file shows (pretty printed for more clarity):

```xml
<testsuite tests="2" errors="2" failures="0" hostname="xxx" name="unittest.case._SubTest" time="0.0002923011779785156" timestamp="2026-04-10T19:12:10.863792">
  <properties />
  <testcase classname="unittest.case._SubTest" name="ek [0]" time="0.00014638900756835938">
    <error message="name \'y\' is not defined" type="&lt;class \'NameError\'&gt;">name \'y\' is not defined

  File "/tmp/tmpr52lsmpk/testrunner-ex/sample2/sampletests_e.py", line 33, in sample2.sampletests_e.eek
    &gt;&gt;&gt; f()
  File "&lt;doctest sample2.sampletests_e.eek[0]&gt;", line 1, in &lt;module&gt;
  File "/tmp/tmpr52lsmpk/testrunner-ex/sample2/sampletests_e.py", line 20, in f
    g()
    ~^^
  File "/tmp/tmpr52lsmpk/testrunner-ex/sample2/sampletests_e.py", line 27, in g
    x = y + 1  # noqa: F821
        ^
</error>
  </testcase>
  <testcase classname="unittest.case._SubTest" name="" time="0.00014591217041015625">
    <error message="name \'x\' is not defined" type="&lt;class \'NameError\'&gt;">name \'x\' is not defined

  File "/tmp/tmpr52lsmpk/testrunner-ex/sample2/e.rst", line 4, in e.rst
    &gt;&gt;&gt; f()
  File "&lt;doctest e.rst[1]&gt;", line 1, in &lt;module&gt;
  File "&lt;doctest e.rst[0]&gt;", line 2, in f
</error>
  </testcase>
  <system-out />
  <system-err />
</testsuite>
```

Which is the traceback from [`e.rst`](https://github.com/zopefoundation/zope.testrunner/blob/master/src/zope/testrunner/tests/testrunner-ex/sample2/e.rst) and [`sampletests_e.py`](https://github.com/zopefoundation/zope.testrunner/blob/master/src/zope/testrunner/tests/testrunner-ex/sample2/sampletests_e.py).

Looking at CPython, I guess that this comes from [this commit](https://github.com/python/cpython/commit/393773ae8763202ecf7afff81c4d57bd37c62ff6#diff-c6fe1ffe930def48a6adf1fa99b974737bac586fdacccacd1474e7b2f11370eb). There are only 2 commits on `Lib/unittest/case.py` that are only 3.15+ and the most recent one is only adding an extra parameter, so the other one might be.

Back to the topic though, I would say that the fix should be rather to increase the number to `107` and make it conditional on the python version, as that new file has valuable information IMHO, rather than removing them.

Subtests are an important, albeit not too much used?, feature of `unittest.case`.

I might be completely wrong though 😅

frenzymadness:
Upated and force-pushed.

icemac (approved):
Thank you for your PR, I slightly improved it to support our tooling for removing it when it is no longer needed.

